Issue #3, opened 2024-11-27
Investigate search page hooks


malberts:
Areas that need to be intercepted:
* top of search form (`.mw-search-form-wrapper`) - for adding "item type" filter
* side of search results - for facets
* individual search results - to rewrite the link

AC:
* Add stub implementation for each hook

PR #6, opened 2024-11-27
Add search hooks

malberts:
For #3 

Adds hooks for:
* changing search result title/link
* adding facets container

----

This does not add a hook for modifying the form part of the search page. Whether we want to do this depends on how item type selection should happen. However, if we do need to modify the form, here's what I found so far:
* [SpecialSearchProfileForm hook](https://www.mediawiki.org/wiki/Manual:Hooks/SpecialSearchProfileForm) can add HTML, but not when using the Advanced tab
* [Extension:AdvancedSearch](https://www.mediawiki.org/wiki/Extension:AdvancedSearch) builds a custom form in Javascript

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,8 +101,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - mw: 'REL1_39'
-            php: '8.2'
+          - mw: 'REL1_43'
+            php: '8.3'
 
     runs-on: ubuntu-latest
 
--- a/extension.json
+++ b/extension.json
@@ -36,6 +36,8 @@
 	},
 
 	"Hooks": {
+		"ShowSearchHitTitle": "ProfessionalWiki\\WikibaseFacetedSearch\\EntryPoints\\WikibaseFacetedSearchHooks::onShowSearchHitTitle",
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -101,8 +101,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - mw: 'REL1_39'
-            php: '8.2'
+          - mw: 'REL1_43'
+            php: '8.3'
```
malberts, line 105: To support MW 1.39 we need to use the older class aliases. PHPStan apparently cannot figure them out on its own and [requires additional config](https://phpstan.org/user-guide/discovering-symbols#class-aliases).

Quickest solution for now is to run PHPStan CI on 1.43 which matches the baseline's ignored errors.

extension.json
```diff
@@ -47,6 +49,15 @@
...
 
 	"ResourceModules": {
+		"ext.wikibase.facetedsearch.styles": {
+			"styles": [
+				"ext.wikibase.facetedsearch.less"
+			],
+			"targets": [
+				"mobile",
+				"desktop"
+			]
```
JeroenDeDauw, line 59: Arn't these the defaults?

Issue #4, opened 2024-11-27
Rewrite search result link to page in sitelinks


malberts:
Partial follow-up to #3 

Wikibase items in the search results need to link to a wiki page based on a specific sitelink.

JeroenDeDauw:
We will need to specify the "site code" to Wikibase, which we need to have in our configuration.

I expect something like this to work

```php
Title::newFromText( SiteLinkLookup::getSiteLinksForItem( new ItemId( 'Q123' ) )[$siteCodeFromOurConfig]->getPageName() )
```

* Need to get SiteLinkLookup instance
* Need to ensure the correct namesaces title is constructed when getPageName returns `Person:172415`

PR #24, opened 2024-11-30
Rewrite item link with page sitelink

malberts:
For #4


phpstan.neon
```diff
@@ -9,4 +9,8 @@ parameters:
 		- ../../includes
 		- ../../tests/phpunit
 		- ../../vendor
+		- ../../extensions/CirrusSearch
 		- ../../extensions/Wikibase
+		- ../../extensions/WikibaseCirrusSearch
+	bootstrapFiles:
+		- ../../includes/AutoLoader.php
```
malberts, line 16: This is for PHPStan to understand `class_alias()`.

src/EntryPoints/WikibaseFacetedSearchHooks.php
```diff
@@ -26,11 +32,47 @@ public static function onShowSearchHitTitle(
...
+				$titleSnippet,
+				$result,
+				$terms,
+				$specialSearch,
+				$query,
+				$attributes
+			);
+		} else {
+			self::runHtmlPageLinkRendererEndHook( $title, $titleSnippet, $attributes );
+		}
```
malberts, line 59: When replacing $title, the link gets the text from the new Title object. We want to keep the original Item title, so we have to let the original link text builders run.

malberts:
All 4 scenarios work after rewriting this.

1. When searching with a mix of namespaces (Item and Main):
image

2. With just Item namespace:
image
 
3. Search for Item ID with mix of namespaces:
image

4. Search for Item ID with just Item namespace:
image


The default, additional language text also works:
image

src/EntryPoints/WikibaseFacetedSearchHooks.php
```diff
@@ -26,11 +35,93 @@ public static function onShowSearchHitTitle(
...
+
+		$pageTitle = self::getItemPage( $itemId );
+
+		if ( $pageTitle === null ) {
 			return;
 		}
 
-		// TODO: get item site link and replace $title
+		$title = $pageTitle;
```
malberts, line 50: For EntityResult results we just need to point to the new page, since the correct $titleSnippet gets set in `CirrusShowSearchHitHandler::onShowSearchHitTitle()`.

src/EntryPoints/WikibaseFacetedSearchHooks.php
```diff
@@ -26,11 +35,93 @@ public static function onShowSearchHitTitle(
...
+		if ( !( $result instanceof EntityResult ) ) {
+			self::rewriteLinkForNonEntityResult(
+				self::newLabelDescriptionLookup( $specialSearch->getContext() ),
+				self::newLinkFormatter( $specialSearch->getLanguage() ),
+				$itemId,
+				$titleSnippet,
+				$attributes
+			);
+		}
+	}
```
malberts, line 61: For non-EntityResult there is no $titleSnippet by default, which means the link is going to use the text of the target page title, not the item title.

To get the item title again, this calls a minimal implementation based on `HtmlPageLinkRendererEndHookHandler::internalDoHtmlPageLinkRendererEnd()`.

src/Persistence/SiteLinkItemPageLookup.php
```diff
@@ -0,0 +1,32 @@
...
+		private readonly SiteLinkLookup $siteLinksStore,
+		private readonly string $siteLinkSiteId
+	) {
+	}
+
+	public function getPageTitle( ItemId $itemId ): ?Title {
+		$siteLink = array_filter(
+			$this->siteLinksStore->getSiteLinksForItem( $itemId ),
+			fn( $siteLink ) => $siteLink->getSiteId() === $this->siteLinkSiteId
+		)[0] ?? null;
```
JeroenDeDauw, line 23: The `array_filter()[0] ?? null` could go into its own function

src/EntryPoints/WikibaseFacetedSearchHooks.php
```diff
@@ -26,11 +35,94 @@ public static function onShowSearchHitTitle(
...
+	 */
+	private static function termFallbackToTermData( ?TermFallback $term = null ): ?array {
+		if ( $term ) {
+			return [
+				'value' => $term->getText(),
+				'language' => $term->getActualLanguageCode(),
+			];
+		}
+
+		return null;
```
JeroenDeDauw, line 125: There is a lot of static code here. Likely we can move much to one or more classes that do not bind to globals and can be tested easier

PR #26, opened 2024-11-30
Sitelink refactor

malberts:
* Add ItemPageLookupFactory to allow for non-sitelink page lookups.
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/24#discussion_r1864322169
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/24#discussion_r1864323057

### DIFF
--- a/src/Persistence/ItemPageLookupFactory.php
+++ b/src/Persistence/ItemPageLookupFactory.php
@@ -1,0 +1,27 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Persistence;
+
+use ProfessionalWiki\WikibaseFacetedSearch\Application\Config;
+use Wikibase\Repo\WikibaseRepo;
+
+class ItemPageLookupFactory {
+
+	public function __construct(
+		private readonly Config $config
+	) {
+	}
+
+	public function newItemPageLookup(): ItemPageLookup {
+		if ( $this->config->linkTargetSitelinkSiteId !== null ) {
+			return new SiteLinkItemPageLookup(
+				WikibaseRepo::getStore()->newSiteLinkStore(),
+				$this->config->linkTargetSitelinkSiteId
+			);
+		}
+
+		return new NullItemPageLookup();
+	}
+}
--- a/src/Persistence/NullItemPageLookup.php
+++ b/src/Persistence/NullItemPageLookup.php
@@ -1,0 +1,16 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Persistence;
+
+use Title;
+use Wikibase\DataModel\Entity\ItemId;
+
+class NullItemPageLookup implements ItemPageLookup {
+
+	public function getPageTitle( ItemId $itemId ): ?Title {
+		return null;
+	}
+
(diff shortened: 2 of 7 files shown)


tests/Persistence/NullItemPageLookupTest.php
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * @covers \ProfessionalWiki\WikibaseFacetedSearch\Persistence\NullItemPageLookup
+ */
+class NullItemPageLookupTest extends TestCase {
+
+	public function testReturnsNull(): void {
+		$this->assertNull(
+			( new NullItemPageLookup() )->getPageTitle( new ItemId( 'Q42' ) )
+		);
```
JeroenDeDauw, line 19: Someone wants their 100% code coverage :D

src/Persistence/ItemPageLookupFactory.php
```diff
@@ -0,0 +1,27 @@
...
+		if ( $this->config->linkTargetSitelinkSiteId !== null ) {
+			return new SiteLinkItemPageLookup(
+				WikibaseRepo::getStore()->newSiteLinkStore(),
+				$this->config->linkTargetSitelinkSiteId
+			);
+		}
+
+		return new NullItemPageLookup();
+	}
+}
```
JeroenDeDauw, line 27: What motivated you to put this factory logic into a dedicated class? (I'm not objecting)

src/Persistence/NullItemPageLookup.php
```diff
@@ -0,0 +1,16 @@
...
+namespace ProfessionalWiki\WikibaseFacetedSearch\Persistence;
+
+use Title;
+use Wikibase\DataModel\Entity\ItemId;
+
+class NullItemPageLookup implements ItemPageLookup {
+
+	public function getPageTitle( ItemId $itemId ): ?Title {
+		return null;
+	}
```
JeroenDeDauw, line 14: I wonder if the call to `rewriteLinkForNonEntityResult` in the hooks file is being skipped over in cases it should not when we always return null.

JeroenDeDauw (approved):

src/Persistence/ItemPageLookupFactory.php
malberts: I figured it would be slightly more testable than being directly in the extension factory (where we don't typically expect, and test, logic). But then again it's also slightly YAGNI since there are no other Lookups at the moment.

src/Persistence/NullItemPageLookup.php
malberts: `rewriteLinkForNonEntityResult` basically reimplements this: https://github.com/wikimedia/mediawiki-extensions-Wikibase/blob/3f636e441cf070234bfcd4f217603d2e2a0be28a/repo/includes/Hooks/HtmlPageLinkRendererEndHookHandler.php#L321-L327

When you override `$title` without `$titleSnippet` it ends up using `$title`'s page title for the link. So to get the original link text (derived from WB item) back, we need to trigger the original link text building. I couldn't get it to work by manually triggering the entire hook that calls that code.

So if `$pageTitle` is null, then we shouldn't do anything further, because the original entity link would already be correct as per the link building of Wikibase/WikibaseCirrusSearch.

PR #28, opened 2024-12-06
Add elastica dependency to composer.json

JeroenDeDauw:


composer.json
```diff
@@ -24,7 +24,8 @@
 	"require": {
 		"php": ">=8.1",
 		"composer/installers": "^2|^1.0.1",
-		"opis/json-schema": "^2.3.0"
+		"opis/json-schema": "^2.3.0",
+		"ruflin/elastica":  "^7.3.0"
```
malberts, line 28: If we're going to support MW 1.39 then we have to lower this to: "^7.1.5"
https://github.com/wikimedia/mediawiki-extensions-Elastica/blob/13b91fc23a8e1b12a128bb129610820df91127c9/composer.json#L15

Issue #9, opened 2024-11-28
Facets UI: Add UI


JeroenDeDauw:
AC:
* Facets are shown
* Do or create task for: make UI look nice

Not AC:
* Real data (https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/10)
* Real config
* Ability to filter search results by clicking facet (https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/11)

Part of
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/8

Issue #32, opened 2024-12-09
Add facet type for years


JeroenDeDauw:
Let's do this one near the end of the development

Initial through was a "slider" with two ticks for min and max. However, having a min and max numeric inputs has these advantages:
* Support for open ranges (without having to implement something special). Example: "all papers published after 2020"
* No need to determine what the min and max values on the "slider" should be when building the facet UI

Need a product decision on the above

Open questions:
* What granularity for the data in elastic
* What granularity in the UI. Do we allow configuring this?

Preliminary implementation of the Elastic index done by:
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/39

JeroenDeDauw:
Decision: min and max inputs (allowing for open ranges)

PR #34, opened 2024-12-11
Add Mustache-based FacetUiBuilder

malberts:
For #9 

Only minimally styled.

Screenshot_20241211_023431

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
 		"php": ">=8.1",
 		"composer/installers": "^2|^1.0.1",
 		"opis/json-schema": "^2.3.0",
-		"ruflin/elastica":  "^7.1.0"
+		"ruflin/elastica":  "^7.1.0",
+		"psr/log": "^1.1.4"
 	},
 	"require-dev": {
 		"phpstan/phpstan": "^2.0.1",
--- a/resources/ext.wikibase.facetedsearch.less
+++ b/resources/ext.wikibase.facetedsearch.less
@@ -1,6 +1,31 @@
 @import 'mediawiki.skin.variables.less';
 
-.wikibase-faceted-search__facets {
-	border: @border-width-base @border-style-base @border-color-base;
-	padding: @spacing-100;
+.wikibase-faceted-search {
+	// Common
+	&__facets {
+		border: @border-width-base @border-style-base @border-color-base;
+		padding: @spacing-100;
+	}
+
+	&__facet {
+		border: @border-width-base @border-style-base @border-color-base;
+		padding: @spacing-100;
+		margin-bottom: @spacing-100;
+	}
+
+	&__facet-label {
+		font-weight: bold;
+	}
+
+	// List facet
+	&__facet-items {
+		list-style: none;
+		margin: 0;
+	}
+
+	&__facet-item {
+		&--selected {
+			font-weight: bold;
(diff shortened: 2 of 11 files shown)


templates/RangeFacet.mustache
malberts, line 1: Looks like this is a placeholder for https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/32

resources/ext.wikibase.facetedsearch.less
malberts, line 1: Using CSS BEM.

composer.json
```diff
@@ -25,7 +25,8 @@
 		"php": ">=8.1",
 		"composer/installers": "^2|^1.0.1",
 		"opis/json-schema": "^2.3.0",
-		"ruflin/elastica":  "^7.1.0"
+		"ruflin/elastica":  "^7.1.0",
+		"psr/log": "^1.1.4"
```
malberts, line 29: The elastica import was pulling in psr/log 3.x. Meanwhile, MW requires that to be at 1.1.4, due to method signature imcompatibilities. This was causing the previous CI failures.

JeroenDeDauw: So the reason you need a new desk is that MediaWiki devs can't define their dependencies?

Bit of a weird approach in MediaWiki. Either you get inappropriate pinning like `5.21.6.47.1.9.52.3`, or the constraint is missing.

resources/ext.wikibase.facetedsearch.less
JeroenDeDauw: BEM?

templates/RangeFacet.mustache
```diff
@@ -0,0 +1,6 @@
+<div>
+	{{msg-min}}: <input name="min" type="text" value="{{min}}" />
+</div>
+<div>
+	{{msg-max}}: <input name="max" type="text" value="{{max}}" />
```
JeroenDeDauw, line 5: Ah, so these min and max are `currentMin` and `currentMax`. That's not obvious in the caller code, where I was thinking they represent `rangeMin` and `rangeMax`. I'm happy these are the current values, since it's not clear we will need the boundaries of the range.

composer.json
malberts: MW actually pins it as "1.1.4", but we're getting composer dependencies in both MW and separately in the extension. MW's `vendor` has the right version. I think something weird was going on with PHPStan using (some of?) the extension's vendors. So it ends up analysing MW code using a wrong version of psr/log, because ruflin/elastica allows 1|2|3.

templates/RangeFacet.mustache
malberts: Boundaries might be a UX requirement if we do the slider, otherwise there's no limit to where you slide. But otherwise as far as the query goes, there is no technical requirement for boundaries. If `rangeMax` was calculated as 1000 by ElasticSearch (i.e. the matching items only go up to 1000) and you enter 2000, then there should be no additional matching items anyway.

composer.json
JeroenDeDauw: We might be installing PHPStan in the extension itself, ignoring MW dependencies. Such an approach is often needed to avoid conflicts with those MediaWiki pins or to be able to use PHPUnit that is not from 2007.

Issue #13, opened 2024-11-28
Add Facets config


JeroenDeDauw:
Depends on
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/12

AC:
* Can define per item type ("instance of"), where an item type is defined by a Q-id
    * Can specify the P-id of the item type property. (Unless this is already defined in Wikibase somewhere)
* Can order facets
* Can have multiple facets for one property (P-id). Example: "boolean Has Author filter" and "Author list filter"
* Can specify config for a facet beyond the property id. Example: "boolean Has Author filter" vs "Author list filter"

Ideally also immediately document, at least when there is no high likleyhood of imminent change

---

Possible approach:

```json
{
	"facets": [
		{
			"property": "P1",
			"type": "requireValue"
		},
		{
			"property": "P1",
			"type": "valueList"
		},
		{
			"property": "P2",
			"type": "slider"
		}
	]
}

```

PR #35, opened 2024-12-11
Add FacetConfig

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/13

Config page: 
image

The documentation likely needs some follow-up expansion.

### DIFF
--- a/resources/ext.wikibase.facetedsearch.docs.less
+++ b/resources/ext.wikibase.facetedsearch.docs.less
@@ -1,0 +1,3 @@
+#Documentation pre {
+	tab-size: 4;
+}
--- a/src/Application/FacetConfig.php
+++ b/src/Application/FacetConfig.php
@@ -12,7 +12,7 @@ class FacetConfig {
 	public function __construct(
 		public readonly ItemId $itemId,
 		public readonly PropertyId $propertyId,
-		public readonly string $type
+		public readonly FacetType $type
 	) {
 	}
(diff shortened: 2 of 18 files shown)


src/Application/FacetConfig.php
```diff
@@ -0,0 +1,19 @@
...
+
+use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Entity\PropertyId;
+
+class FacetConfig {
+
+	public function __construct(
+		public readonly ItemId $itemId,
+		public readonly PropertyId $propertyId,
+		public readonly string $type
```
malberts, line 15: This should be something more specific than a string. I don't think we need a full widget registry, and instead an enum will be fine.

resources/ext.wikibase.facetedsearch.docs.less
```diff
@@ -0,0 +1,3 @@
+#Documentation pre {
+	tab-size: 4;
```
malberts, line 2: I added this because the default 8 space indent looks bad on the config page.

src/Application/FacetConfig.php
JeroenDeDauw: Agreed, we can use an Enum. No need to support adding widget types via extensions to WBFS

JeroenDeDauw (approved):

Issue #31, opened 2024-12-09
Parse facet filter info out of search string


JeroenDeDauw:
Needed by
* The code building the facets, so it knows which values to mark as already selected 
* Facets when they need to update the overall search query string, so they know what part to modify: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/11

JeroenDeDauw:
@malberts what is the state of this? I know there are unknowns around ranges, but the most common filters we could already do, unblocking some dependent tasks.

JeroenDeDauw:
Done with https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/67

PR #41, opened 2024-12-15
Add TODOs in FacetUiBuilder and make facet config accessible

JeroenDeDauw:


src/Application/Config.php
```diff
@@ -4,14 +4,16 @@
...
 use Wikibase\DataModel\Entity\PropertyId;
 
 class Config {
 
 	public function __construct(
 		public readonly ?string $linkTargetSitelinkSiteId = null,
-		public readonly ?PropertyId $instanceOfId = null,
-		public readonly ?FacetConfigList $facets = null
+		private readonly ?PropertyId $instanceOfId = null,
+		private readonly ?FacetConfigList $facets = null
```
JeroenDeDauw, line 16: I switched to getters, since for access in our application, after the config has been figured out, these values are no longer nullable.

src/Application/ConfigLookup.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
```
JeroenDeDauw, line 5: The interfaces go within Application/, their persistence-mechanism-specific implementations go into Persistence/. 

(It is possible to have interfaces end up in Persistence/ if they are used by a class in there, in which case they won't be used in Application/)

JeroenDeDauw: The interfaces form the architectural boundary

tests/Valid.php
```diff
@@ -49,10 +49,10 @@ public static function config(): Config {
 			linkTargetSitelinkSiteId: 'enwiki',
 			instanceOfId: new NumericPropertyId( 'P42' ),
 			facets: new FacetConfigList(
-				new FacetConfig( new ItemId( 'Q100' ), new NumericPropertyId( 'P100' ), FacetType::BOOLEAN ),
```
JeroenDeDauw, line 52: Q100, Q200, P100, P200 are similar. Nice to make values more distinct. Q100 Q200 P1 P2 is less likely to result in confusion and is easier to keep in your short term memory

src/Application/Config.php
```diff
@@ -23,4 +25,26 @@ public function combine( Config $config ): self {
...
+		}
+
+		return $this->instanceOfId;
+	}
+
+	/**
+	 * TODO: make private?
+	 */
+	public function getFacets(): FacetConfigList {
+		return $this->facets ?? new FacetConfigList();
```
JeroenDeDauw, line 40: I ended up adding the below `getFacetConfigForInstanceType`, which is what made me put the TODO on this function.

src/Application/FacetConfigList.php
```diff
@@ -20,14 +20,14 @@ public function __construct( FacetConfig ...$facets ) {
 	}
 
 	private function addFacetConfig( FacetConfig $facetConfig ): void {
-		$this->facetsPerItemId[$facetConfig->itemId->getSerialization()][] = $facetConfig;
+		$this->facetsPerInstanceType[$facetConfig->instanceTypeId->getSerialization()][] = $facetConfig;
```
JeroenDeDauw, line 23: There is arguably some technical debt here, but it's fully contained, so it does not matter.

Two guidelines violated here:
* Compute things when you need them, don't put them in place just in case you need them later
* Do not do work in the constructor

malberts: The original idea was that we'd need to access per-item/instance facet config when building the actual facets. So to build FacetConfigList I had to decide how to keep track of the configs and that was the first use case. However, in https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/31 I'm going to need the facet configs but without caring which items/instances they are for - which results in doing `array_values()` on this item-indexed array. So it ends up requiring additional processing either way.

Probably we should just put them in a plain list and have something like `getFacetsByInstanceType()` that returns them indexed by item/instance ID.

src/Application/Config.php
malberts: We might need both, or at least another version. See https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/41/files#r1885911926

src/Application/ConfigLookup.php
malberts: Blame WikibaseExport developers!

PR #42, opened 2024-12-15
Add ConfigPageTest

JeroenDeDauw:


tests/Persistence/PageContentConfigLookupTest.php
```diff
@@ -11,6 +11,7 @@
 
 /**
  * @covers \ProfessionalWiki\WikibaseFacetedSearch\Persistence\PageContentConfigLookup
+ * @uses \ProfessionalWiki\WikibaseFacetedSearch\Persistence\PageContentFetcher
```
JeroenDeDauw, line 14: It is common to split code into several classes but only have tests for the higher level one. In such cases it's good to include covers tags for the lower level ones in said test.

tests/Persistence/PageContentConfigLookupTest.php
```diff
@@ -10,11 +10,17 @@
...
  * @group Database
+ * @covers \ProfessionalWiki\WikibaseFacetedSearch\Persistence\PageContentConfigLookup
+ * @covers \ProfessionalWiki\WikibaseFacetedSearch\Persistence\PageContentFetcher
  */
 class PageContentConfigLookupTest extends WikibaseFacetedSearchIntegrationTest {
 
+	public function tearDown(): void {
+		$this->deleteConfigPage();
+		parent::tearDown();
+	}
```
JeroenDeDauw, line 22: This test class was leaking state despite being in `@group Database`

Issue #38, opened 2024-12-13
Add search keyword for facets


malberts:
Not sure we want to do this yet

Part of 
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/32

`haswbstatement` supports only "equals" queries.

A new search keyword is necessary to generate the ElasticSearch query for other operators. Similar to `wbstatementquantity`, but for "Quantity" and "Point In Time" statement _values_.

AC:
* [x] Add keyword
* [ ] Add keyword pattern matching
* [ ] Generate Elastic query

JeroenDeDauw:
Remaining work:
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/53

PR #44, opened 2024-12-17
Add new keyword placeholder

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/38

Doesn't do anything yet.

I went with `haswbfacet` because:
* `haswb...` is the pattern used in WikibaseCirrusSearch
* `facet`, in general, refers to a specific aspect an object. This is reflected in how our new keyword is going to support a subset of `haswbstatement`, specifically limiting keyword usage to one facet (property) at a time.

I suppose we can bikeshed this later.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -14,3 +14,6 @@ parameters:
 		- ../../extensions/WikibaseCirrusSearch
 	bootstrapFiles:
 		- ../../includes/AutoLoader.php
+	excludePaths:
+		- src/EntryPoints/WikibaseFacetedSearchHooks.php
+		- src/Persistence/Search/Query/HasWbFacetFeature.php
--- a/tests/Persistence/Search/Query/HasWbFacetFeatureTest.php
+++ b/tests/Persistence/Search/Query/HasWbFacetFeatureTest.php
@@ -1,0 +1,48 @@
+<?php
+
+declare( strict_types = 1 );
+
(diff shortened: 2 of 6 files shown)


phpstan.neon
```diff
@@ -14,3 +14,5 @@ parameters:
 		- ../../extensions/WikibaseCirrusSearch
 	bootstrapFiles:
 		- ../../includes/AutoLoader.php
+	excludePaths:
+		- src/Persistence/Search/Query/HasWbFacetFeature.php
```
malberts, line 18: This is unfortunate. PHPStan cannot find CirrusSearch classes, because there's no autoloading in their [composer.json](https://github.com/wikimedia/mediawiki-extensions-CirrusSearch/blob/master/composer.json).

I verified that if I manually add
```json
	"autoload": {
		"psr-4": {
			"CirrusSearch\\": "includes/"
		}
	}
```
and update composer, then PHPStan is happy.

So basically, either we:
* have no PHPStan on code touching CirrusSearch; or
* patch CirrusSearch locally and in CI

tests/Persistence/Search/Query/HasWbFacetFeatureTest.php
malberts, line 1: Most of the code in here comes from HasWbStatementFeatureTest (WikibaseCirrusSearch) just have a basic test running.

PR #45, opened 2024-12-23
Give $itemType to FacetUiBuilder

JeroenDeDauw:


src/Presentation/FacetUiBuilder.php
```diff
@@ -19,9 +19,8 @@ public function __construct(
 
 	// TODO: parameter or constructor argument: values and counts (from https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/23)
 	// TODO: parameter: selected values (from QueryStringParser https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/31)
-	// TODO: parameter: instance of item ID
-	public function createHtml(): string {
-		$this->config->getFacetConfigForInstanceType( new ItemId( 'Q1' ) ); // TODO: real id & use
+	public function createHtml( ItemId $itemType ): string {
```
malberts, line 22: The name feels a little confusing. Item type

src/Presentation/FacetUiBuilder.php
```diff
@@ -19,9 +19,8 @@ public function __construct(
...
+		$this->config->getFacetConfigForInstanceType( $itemType );
```
malberts, line 23: I think this might not be correct. A search result can contain multiple different item instance types. Unless the user filtered on a specific type - which I believe is optional. So we would likely have to use a list here.

JeroenDeDauw: I don't see the issue. Yes, search results can contain items of multiple types, so what? This code is not about rendering search results.

We need the item type info because if the user filtered on Publications, we need to show the facets for Publications. If they did not filter, we do not show any facets, beyond the item type selector. This item type selector would only allow selecting one value.

Theoretically the user could do something else with the query string, but we can just bail out, or pick the first type. Something the query string parsing code can decide. With current approach, the HTML builder here only cares about what type facets should be shown for. Probably we should make the ID nullable so it can also handle the "no item type is selected yet, so only show the item type selector" case.

malberts: > If they did not filter, we do not show any facets, beyond the item type selector. 

In that case it is fine. I was under the impression the type selector would be optional, in which case we'd show facet filters for all the types in the search

Issue #49, opened 2024-12-29
Create Elasticsearch fields


malberts:
AC:
* Add dedicated Elastic field per configured facet
  * example: `wbfs_P42`

Related:
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/39

PR #50, opened 2024-12-29
Add initial search fields

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/49

Creates an Elasticsearch field for each configured facet.

Example in Kibana:
Screenshot_20241229_231759

### DIFF
--- a/src/Persistence/Search/FacetSearchIndexFieldsBuilder.php
+++ b/src/Persistence/Search/FacetSearchIndexFieldsBuilder.php
@@ -1,0 +1,59 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Persistence\Search;
+
+use Exception;
+use ProfessionalWiki\WikibaseFacetedSearch\Application\Config;
+use SearchEngine;
+use SearchIndexField;
+use Wikibase\DataModel\Services\Lookup\PropertyDataTypeLookup;
+
+class FacetSearchIndexFieldsBuilder {
+
+	public function __construct(
+		private readonly SearchEngine $engine,
+		private readonly Config $config,
+		private readonly PropertyDataTypeLookup $dataTypeLookup
+	) {
+	}
+
+	/**
+	 * @return array<string, SearchIndexField>
+	 */
+	public function createFields(): array {
+		$fields = [];
+
+		foreach ( $this->config->getFacets()->asArray() as $facetConfig ) {
+			$name = 'wbfs_' . $facetConfig->propertyId->getSerialization();
+
+			try {
+				$dataTypeId = $this->dataTypeLookup->getDataTypeIdForProperty( $facetConfig->propertyId );
+			} catch ( Exception $e ) {
+				continue;
+			}
+
+			$fieldType = $this->getFieldTypeForDataTypeId( $dataTypeId );
+
+			if ( $fieldType === null ) {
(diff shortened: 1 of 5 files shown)


src/Persistence/Search/FacetSearchIndexFieldsBuilder.php
```diff
@@ -0,0 +1,59 @@
...
+				continue;
+			}
+
+			$fieldType = $this->getFieldTypeForDataTypeId( $dataTypeId );
+
+			if ( $fieldType === null ) {
+				continue;
+			}
+
+			$fields[$name] = $this->engine->makeSearchFieldMapping( $name, $fieldType );
```
JeroenDeDauw, line 43: `$name` can be computed closer to where it is used and the details of said computation could go into a dedicated private method.

src/Persistence/Search/FacetSearchIndexFieldsBuilder.php
```diff
@@ -0,0 +1,59 @@
...
+	 */
+	public function createFields(): array {
+		$fields = [];
+
+		foreach ( $this->config->getFacets()->asArray() as $facetConfig ) {
+			$name = 'wbfs_' . $facetConfig->propertyId->getSerialization();
+
+			try {
+				$dataTypeId = $this->dataTypeLookup->getDataTypeIdForProperty( $facetConfig->propertyId );
+			} catch ( Exception $e ) {
```
JeroenDeDauw, line 33: I suspect we can do 

`} catch ( Exception ) {`

given we use modern PHP

src/Persistence/Search/FacetSearchIndexFieldsBuilder.php
```diff
@@ -0,0 +1,59 @@
...
+				continue;
+			}
+
+			$fieldType = $this->getFieldTypeForDataTypeId( $dataTypeId );
```
JeroenDeDauw, line 37: This line and the above try catch could also be in its own method: getFieldTypeForPropertyId

JeroenDeDauw (approved):

src/Persistence/Search/FacetSearchIndexFieldsBuilder.php
```diff
@@ -0,0 +1,59 @@
...
+		return $fields;
+	}
+
+	private function getFieldTypeForDataTypeId( string $dataTypeId ): ?string {
+		return match ( $dataTypeId ) {
+			'quantity' => SearchIndexField::INDEX_TYPE_NUMBER,
+			'string' => SearchIndexField::INDEX_TYPE_KEYWORD,
+			'time' => SearchIndexField::INDEX_TYPE_DATETIME,
+			'wikibase-item' => SearchIndexField::INDEX_TYPE_KEYWORD,
+			default => null
```
JeroenDeDauw, line 55: I did some double checking about this approach, in particular if it's better to use Wikibase's internal type, and it seems good as it is.

Here is what I did in Semantic Wikibase, which has a fuller list of types: https://github.com/ProfessionalWiki/SemanticWikibase/blob/master/src/Translation/PropertyTypeTranslator.php#L31

JeroenDeDauw: Which makes me think we could go further than extracting a getFieldTypeForPropertyId method as suggested above, and instead make that a service.

PR #51, opened 2024-12-29
Index values

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/56

Example of an Item's statements getting indexed on the page, via sitelink:
image

### DIFF
--- a/src/Application/DataValueTranslator.php
+++ b/src/Application/DataValueTranslator.php
@@ -1,0 +1,35 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
+
+use DataValues\DataValue;
+use DataValues\StringValue;
+use DataValues\TimeValue;
+use DataValues\UnboundedQuantityValue;
+use Wikibase\DataModel\Entity\EntityIdValue;
+
+class DataValueTranslator {
+
+	public function translate( DataValue $dataValue ): mixed {
+		if ( $dataValue instanceof UnboundedQuantityValue ) {
+			return $dataValue->getAmount()->getValueFloat();
+		}
+
+		if ( $dataValue instanceof StringValue ) {
+			return $dataValue->getValue();
+		}
+
+		if ( $dataValue instanceof TimeValue ) {
+			return ltrim( $dataValue->getTime(), '+-' );
+		}
+
+		if ( $dataValue instanceof EntityIdValue ) {
+			return $dataValue->getEntityId()->getSerialization();
+		}
+
+		return null;
+	}
+
+}
--- a/src/Application/StatementListTranslator.php
+++ b/src/Application/StatementListTranslator.php
@@ -1,0 +1,75 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
+
+use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Statement\StatementList;
+
+class StatementListTranslator {
+
+	public function __construct(
+		private readonly StatementTranslator $statementTranslator,
+		private readonly InstanceTypeExtractor $instanceTypeExtractor,
+		private readonly Config $config
+	) {
+	}
+
+	/**
+	 * @return array<string, mixed>
+	 */
+	public function translateStatements( StatementList $statements ): array {
+		$instanceTypeId = $this->instanceTypeExtractor->getInstanceTypeId( $statements );
+
+		if ( $instanceTypeId === null ) {
+			return [];
+		}
+
+		$propertyIds = $this->getInstanceTypePropertyIds( $instanceTypeId );
+
+		$values = [];
+
+		foreach ( $propertyIds as $propertyId ) {
+			$values[$this->getFieldName( $propertyId )] = $this->translatePropertyStatements( $statements, $propertyId );
+		}
+
+		return $values;
+	}
+
+	/**
+	 * @return PropertyId[]
+	 */
+	private function getInstanceTypePropertyIds( ItemId $instanceTypeId ): array {
+		return array_values(
+			array_map(
+				fn( FacetConfig $config ) => $config->propertyId,
+				$this->config->getFacetConfigForInstanceType( $instanceTypeId )
+			)
+		);
+	}
+
+	private function translatePropertyStatements( StatementList $statements, PropertyId $propertyId ): array {
+		$values = [];
+		$propertyStatements = $statements->getByPropertyId( $propertyId )->getBestStatements()->toArray();
+
+		foreach ( $propertyStatements as $statement ) {
+			$value = $this->statementTranslator->statementToSearchData( $statement );
+
+			if ( $value === null ) {
+				continue;
+			}
+
+			$values[] = $value;
+		}
+
+		// TODO: should empty field (all statements removed) be [] or null?
+		return $values;
+	}
+
+	private function getFieldName( PropertyId $propertyId ): string {
+		return 'wbfs_' . $propertyId->getSerialization();
+	}
+
+}
--- a/src/EntryPoints/WikibaseFacetedSearchHooks.php
+++ b/src/EntryPoints/WikibaseFacetedSearchHooks.php
@@ -7,12 +7,14 @@ namespace ProfessionalWiki\WikibaseFacetedSearch\EntryPoints;
 use CirrusSearch\CirrusSearch;
 use CirrusSearch\Query\KeywordFeature;
 use CirrusSearch\SearchConfig;
+use ContentHandler;
 use EditPage;
 use Html;
 use HtmlArmor;
 use IContextSource;
 use Language;
 use OutputPage;
+use ParserOutput;
 use ProfessionalWiki\WikibaseFacetedSearch\Persistence\ConfigJsonValidator;
 use ProfessionalWiki\WikibaseFacetedSearch\Persistence\Search\Query\HasWbFacetFeature;
 use ProfessionalWiki\WikibaseFacetedSearch\Presentation\ConfigJsonErrorFormatter;
@@ -30,6 +32,7 @@ use Wikibase\Lib\Store\LanguageFallbackLabelDescriptionLookup;
 use Wikibase\Repo\Hooks\Formatters\EntityLinkFormatter;
 use Wikibase\Repo\WikibaseRepo;
 use Wikibase\Search\Elastic\EntityResult;
+use WikiPage;
 
 class WikibaseFacetedSearchHooks {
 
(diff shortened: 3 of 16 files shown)


src/Persistence/Search/FacetSearchIndexFieldsBuilder.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Persistence\Search;
+
+use Exception;
+use ProfessionalWiki\WikibaseFacetedSearch\Application\Config;
+use SearchEngine;
+use SearchIndexField;
+use Wikibase\DataModel\Services\Lookup\PropertyDataTypeLookup;
+
+class FacetSearchIndexFieldsBuilder {
```
JeroenDeDauw, line 13: `Facet` does not seem needed in the name

src/EntryPoints/WikibaseFacetedSearchHooks.php
```diff
@@ -224,4 +241,146 @@ public static function onCirrusSearchAddQueryFeatures( SearchConfig $config, arr
...
+		}
+
+//		print_r("\nentityId:$entityId");
+
+		if ( $entityId === null ) {
+			return;
+		}
+
+		$fields = array_merge( $fields, self::getFieldValues( $entityId ) );
+	}
```
JeroenDeDauw, line 297: Stating the obvious: this logic should be in instantiable class(es).

src/EntryPoints/WikibaseFacetedSearchHooks.php
```diff
@@ -224,4 +241,146 @@ public static function onCirrusSearchAddQueryFeatures( SearchConfig $config, arr
...
+
+	private static function getFieldValues( EntityId $entityId ): array {
+		$entity = WikibaseRepo::getEntityLookup()->getEntity( $entityId );
```
JeroenDeDauw, line 300: I suspect something like this will be nice:

```php
private function getStatements(): StatementList {
		return $this->getStatementsLookup()->getStatements();
	}
	
	private function getStatementsLookup(): StatementsLookup {
		if ( /* sitelink config case */ ) {
			return new SitelinkBasedStatementLookup( /* sitelink config */ );
		}
		
		return new FromPageStatementLookup( $wikiPage /* or similar */ );
	}
```

To encapsulate the logic for getting the statements for both the cases we support (wiki with sitelink config or not).

I switched from Entity to StatementList in the signatures, since we only need the later for our current stuff. Could argue for keeping Entity to potentially support aliases and similar data that one might want facets for, but then again YAGNI.

src/EntryPoints/WikibaseFacetedSearchHooks.php
```diff
@@ -224,4 +241,146 @@ public static function onCirrusSearchAddQueryFeatures( SearchConfig $config, arr
...
+		if ( $dataValue instanceof TimeValue ) {
+			return ltrim( $dataValue->getTime(), '+-' );
+		}
+
+		if ( $dataValue instanceof EntityIdValue ) {
+			return $dataValue->getEntityId()->getSerialization();
+		}
+
+		return null;
+	}
```
JeroenDeDauw, line 384: Good candidate for own service

Similar code in Semantic Wikibase: https://github.com/ProfessionalWiki/SemanticWikibase/blob/master/src/Translation/DataValueTranslator.php#L26

JeroenDeDauw:
@malberts the code I mentioned earlier is `$statements->getBestStatements()` (example in `StatementListTranslator` in SemanticWikibase)

malberts:
PR is getting a bit big, so marking as ready and some TODOs for follow-up.

src/Application/StatementListTranslator.php
```diff
@@ -0,0 +1,75 @@
...
+			array_map(
+				fn( FacetConfig $config ) => $config->propertyId,
+				$this->config->getFacetConfigForInstanceType( $instanceTypeId )
+			)
+		);
+	}
+
+	private function translatePropertyStatements( StatementList $statements, PropertyId $propertyId ): array {
+		$values = [];
+		$propertyStatements = $statements->getByPropertyId( $propertyId )->getBestStatements()->toArray();
```
malberts, line 55: I'm just using `getBestStatements()` (which returns `RANK_PREFERRED` _or else_ `RANK_NORMAL`).
In https://github.com/ProfessionalWiki/SemanticWikibase/blob/master/src/Translation/StatementListTranslator.php#L26 it's also calling `getByRank()` afterwards, which seems to be redundant, since it would just return the same already filtered list.
However, if the idea is to get `RANK_PREFERRED` _and_ `RANK_NORMAL`, then we should remove `getBestStatements()` and only use `getByRank()`.

src/Application/DataValueTranslator.php
```diff
@@ -0,0 +1,35 @@
...
+
+use DataValues\DataValue;
+use DataValues\StringValue;
+use DataValues\TimeValue;
+use DataValues\UnboundedQuantityValue;
+use Wikibase\DataModel\Entity\EntityIdValue;
+
+class DataValueTranslator {
+
+	public function translate( DataValue $dataValue ): mixed {
```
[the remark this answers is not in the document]
JeroenDeDauw: Dunno

src/Application/StatementListTranslator.php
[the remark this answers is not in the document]
JeroenDeDauw: image

src/Application/StatementListTranslator.php
```diff
@@ -0,0 +1,75 @@
...
+		private readonly StatementTranslator $statementTranslator,
+		private readonly InstanceTypeExtractor $instanceTypeExtractor,
+		private readonly Config $config
+	) {
+	}
+
+	/**
+	 * @return array<string, mixed>
+	 */
+	public function translateStatements( StatementList $statements ): array {
```
JeroenDeDauw, line 23: This could use a short comment about what the statements are being translated to

src/Application/StatementListTranslator.php
JeroenDeDauw: Just `getBestStatements` is correct

src/Persistence/FromPageStatementsLookup.php
```diff
@@ -0,0 +1,24 @@
...
+class FromPageStatementsLookup implements StatementsLookup {
+
+	public function getStatements( WikiPage $page ): StatementList {
+		$content = $page->getContent();
+
+		if ( !( $content instanceof ItemContent ) ) {
+			return new StatementList();
+		}
+
+		return $content->getItem()->getStatements();
```
JeroenDeDauw, line 21: After some consideration, I think this code is great.

There is also opportunity to check for `EntityContent` and then see if `getEntity` returns `StatementListProvider`, so we support non-Items. Unclear why we would need that. Too long, don't read: https://www.entropywins.wtf/blog/2017/05/24/generic-entity-handling-code/

src/Persistence/SiteLinkBasedStatementsLookup.php
```diff
@@ -0,0 +1,42 @@
...
+namespace ProfessionalWiki\WikibaseFacetedSearch\Persistence;
+
+use ProfessionalWiki\WikibaseFacetedSearch\Application\StatementsLookup;
+use Wikibase\DataModel\Entity\Item;
+use Wikibase\DataModel\Services\Lookup\EntityLookup;
+use Wikibase\DataModel\Statement\StatementList;
+use Wikibase\Lib\Store\SiteLinkLookup;
+use WikiPage;
+
+class SiteLinkBasedStatementsLookup implements StatementsLookup {
```
JeroenDeDauw, line 14: The newer Wikibase code is using `Sitelink` instead of `SiteLink`, which I suppose is clearer than the original capitalization, making it obvious that it's about the single domain concept rather than "site" and "link"

JeroenDeDauw: Mostly an FYI. We are still consistent with most WB code, so unclear if we should change here

src/WikibaseFacetedSearchExtension.php
```diff
@@ -98,4 +105,40 @@ public function newSearchIndexFieldsBuilder( SearchEngine $engine ): SearchIndex
 		);
 	}
 
+	public function newStatementsLookup(): StatementsLookup {
+		if ( $this->getConfig()->linkTargetSitelinkSiteId !== null ) {
```
JeroenDeDauw, line 109: You can omit the negation

tests/Application/InstanceTypeExtractorTest.php
```diff
@@ -0,0 +1,186 @@
...
+			$this->newStatement( 'P200', new StringValue( 'Bar' ) )
+		);
+
+		$this->assertHasInstanceType( 'Q1', $statements );
+	}
+
+	private function newStatement( string $propertyId, DataValue $value ): Statement {
+		return new Statement(
+			new PropertyValueSnak(
+				new NumericPropertyId( $propertyId ),
```
JeroenDeDauw, line 45: Looks funny that `NumericPropertyId` gets a string

src/Persistence/FromPageStatementsLookup.php
malberts: I tried using Item-specific implementations where they were available. 

While it is indeed technically possible to handle non-Items, my main reasons against doing that:
* Property cannot have a sitelink. This is not technically relevant in this class, but leads to a subset of features compared to Item. Additionally, if the sitelink is configured, none of the Properties will get indexed, since there won't be a linked Page.
* Property is unlikely to have an "instance type" statement. (Property Foo might be an instance of "XYZ Property", but not something like "Person"). Of course, technically it is possible.

If not for above use case, but rather as a more general entity statement lookup - I also don't see a need for that at the moment.

src/Persistence/SiteLinkBasedStatementsLookup.php
malberts: I was wondering about this. We're already using both capitalizations. Maybe we should just rename all our symbols to the lowercase `Sitelink`, and then only use uppercase `SiteLink` when calling or accessing symbols in Wikibase.

tests/Application/InstanceTypeExtractorTest.php
malberts: The constructor takes a string.

src/Persistence/SiteLinkBasedStatementsLookup.php
JeroenDeDauw: Using `Sitelink` everywhere seems good

tests/Application/InstanceTypeExtractorTest.php
JeroenDeDauw: I know. No issue here. Just looks funny

JeroenDeDauw (approved):

src/Application/StatementListTranslator.php
```diff
@@ -0,0 +1,75 @@
...
+		$propertyStatements = $statements->getByPropertyId( $propertyId )->getBestStatements()->toArray();
+
+		foreach ( $propertyStatements as $statement ) {
+			$value = $this->statementTranslator->statementToSearchData( $statement );
+
+			if ( $value === null ) {
+				continue;
+			}
+
+			$values[] = $value;
```
JeroenDeDauw, line 64: Second time today I see this continue construct. Is this AI generated?

To me it seems much more natural to do

```php
if ( $value !== null ) {
    $values[] = $value;
}
```

JeroenDeDauw: End up changing this code since I was looking at it anyway: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/61

malberts: Not AI generated. I suspect when I wrote the original POC there was more logic here and as I cleaned it out it ended up like this.

src/Application/StatementListTranslator.php
JeroenDeDauw: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/62

src/WikibaseFacetedSearchExtension.php
JeroenDeDauw: :scream:

JeroenDeDauw: Flip the code

Issue #37, opened 2024-12-11
Facets UI: Make it look nice


malberts:
Follow-up to https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/9

Rough AC:
* Move the facets into a sidebar (i.e. just move it next to search results)
* Don't over-style
* Should look decent with Vector, Vector 2022, Timeless, Chameleon

Part of
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/8

JeroenDeDauw:
We have a placeholder UI already. It shows under the search results. For instance http://localhost:8484/index.php?search=a&search=a&title=Special%3ASearch&go=Go

![image](https://github.com/user-attachments/assets/aec3b46b-8f35-46f0-97ad-39d0528a19ba)

Relevant code is in the `templates/` directory, and to some degree in `FacetUiBuilder`.

JeroenDeDauw:
Done by 
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/57

Further work
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/60

PR #57, opened 2025-01-02
Use Codex CSS-only components in Facets

alistair3149:
- Use CSS only Codex components when applicable
  - Use Codex Checkbox for List Items
  - Use Codex Radio for Boolean Items
  - Use two inline Codex TextInput for Range Items
  - Use Codex Field for each facet
- Refactor facet to allow mixed item types within a facet. That will be helpful when we want to mix different ways to input value within a facet (e.g. having both checkbox and range)
- Allow facets to be stickied as a sidebar above the desktop breakpoint. The implementation is a bit hacky but it will be followed up later.

image

Related: #37


src/Presentation/FacetUiBuilder.php
```diff
@@ -82,24 +86,28 @@ private function getExampleBooleanItems(): array {
...
 				'selected' => true
 			],
 			[
+				'type' => 'Checkbox',
 				'label' => 'Charlie',
 				'count' => 17,
 				'url' => 'https://example.com/facet/Charlie',
 				'selected' => false
 			],
 			[
```
malberts, line 109: > Refactor facet to allow mixed item types within a facet. That will be helpful when we want to mix different ways to input value within a facet (e.g. having both checkbox and range)

This new item type suggests I can end up with a list facet that mixes checkboxes and radios:

Screenshot_20250103_135636

That seems like it will not work.

The facet type (`FacetType` enum) should dictate how that facet's items/inputs are rendered.

If there is a use case to filter the same facet using different filter types, then that facet will need to be configured twice using 2 different facet types. The only use case I can think of now is something like a "year" facet filtered via list of values and range inputs. However, I don't think that is really a typical scenario.

malberts: The original facet types might not be optimal, though. Perhaps it should rather be something like:
* `single` (radios)
* `multiple` (checkboxes)
* `range` (2 inputs)

In any case, the `FacetType` is what goes into the config (https://github.com/ProfessionalWiki/WikibaseFacetedSearch/blob/master/example.json), and that is what should determine whether the items get rendered as radios/checkboxes/etc.

The example item methods in the current code will eventually be replaced by data based on the current search. However, this data won't include how to render an item. It will just be the data about the available facet items (label, value, count, selected, etc).

malberts:
The new UI looks good. It'll be nice if clicking the radio/checkbox label can also (de)select it.

src/Presentation/FacetUiBuilder.php
JeroenDeDauw: Definitely these:
* `multiple` (checkboxes) (be it for the AND or OR case)
* `range` (2 inputs)

Perhaps
* `single` (radios)
    * we still need a way for users to pick the item type they are searching for, before we can show the facets (which are all item-type-specific). This could be done with a radio select, though is perhaps not relevant for the standard facet code, since it'd be build via different logic.
    * There might be cases where there never is overlap in the data. For instance publication type. This would be fine with a list of checkboxes, though perhaps having radio buttons is slightly better UX. Does not seem like something we need in MVP

src/Presentation/FacetUiBuilder.php
```diff
@@ -109,25 +117,50 @@ private function getExampleListItems(): array {
...
+		foreach ( $items as $item ) {
+			if (
+				!isset( $item['type'] ) ||
+				!is_string( $item['type'] ) ||
+				!strlen( $item['type'] ) > 0
+			) {
+				continue;
+			}
+
+			$html .= $this->parser->processTemplate(
```
JeroenDeDauw, line 158: Would be cleaner if you do

```
if ( $this->hasValidType( $item ) ) {
    $html .= $this->parser->processTemplate();
}
```

JeroenDeDauw: (Sidenote: I suspect that once we finish up the code, we will not have an $items array any more, and have more guarantees about the data we get. Bit hard to see as this class is very WIP)

extension.json
```diff
@@ -69,8 +69,16 @@
...
+			"class": "MediaWiki\\ResourceLoader\\CodexModule",
 			"styles": [
 				"ext.wikibase.facetedsearch.less"
+			],
+			"codexStyleOnly": true,
+			"codexComponents": [
+				"CdxCheckbox",
+				"CdxField",
+				"CdxRadio",
+				"CdxTextInput"
```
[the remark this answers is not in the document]
alistair3149: > Looks like this requires MW 1.42+: https://www.mediawiki.org/wiki/Codex#Usage_without_JavaScript_(CSS-only_Codex_components).
> 
> Would this cause a problem in MW 1.39, or would it just end up loading the JS unnecessarily?

Ah yes this requires 1.42+ as Codex CSS-only component is only available on 1.43. There are no JS loaded, only styles for the Codex components listed.

> for context: while there is no hard requirement in this project to keep compat with 39, we figured we'd keep compat unless doing so turns out to be a pile of work

In that case, would we be able to target 1.43 instead? It is much less troublesome if we can use Codex for UI.

resources/ext.wikibase.facetedsearch.less
```diff
@@ -24,8 +23,62 @@
...
+
+			&::after {
+				content: ')';
+			}
+		}
+	}
 }
+
+// HACK: Creating a multi-column layout
+// TODO: Find a way to put .wikibase-faceted-search inside .searchresults
```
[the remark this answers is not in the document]
alistair3149: I need to look into SearchVue to see how did they do it properly. We can follow up in a separate PR on it

src/Presentation/FacetUiBuilder.php
alistair3149: Yes radio and checkbox are not supposed to be mixed. Originally my idea was to allow different input types within the same facet. For example, we can put checkbox and toggle switch together. That will grant us more flexibility on building and formatting the facet.

extension.json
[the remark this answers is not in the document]
JeroenDeDauw: If you think it is too much of a bother to support MW 39, then let's drop support for it.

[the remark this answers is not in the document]
malberts: Something to weigh up is how much of Codex we really need. We're not building a complicated UI. It is indeed nice to be futureproof with Codex, but the trade-off in this specific case is essentially to drop MW 1.39 support in exchange for styled form inputs.

However, requiring MW 1.43+ would also allow us to remove the use of some deprecated PHP class aliases. So if we decide to drop MW 1.39 we should do that as soon as possible.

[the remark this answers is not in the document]
alistair3149: > If you think it is too much of a bother to support MW 39, then let's drop support for it. Not ideal from a product/adoption perspective, but also not a big deal.

We can backport tokens and re-create the components somewhat like in SMW and SRF, it is viable but it is more complicated and more troublesome to maintain.

> Something to weigh up is how much of Codex we really need. We're not building a complicated UI.

At least for the current PR, we are only using CSS-only Codex components, so we are using Codex in its full extent since we don't have any JS set up yet. While we might not be building complicated UI, Codex offers a hassle-free way to provide UI cohesion with other MW components, and offers in-depth accessibility and interactions which is time-consuming to create from scratch. 

> However, requiring MW 1.43+ would also allow us to remove the use of some deprecated PHP class aliases. So if we decide to drop MW 1.39 we should do that as soon as possible.

That is one of the reason that I suggest targeting 1.43 as well.
1.39 and 1.43 have the most amount of (breaking) changes in history (e.g. [T166010](https://phabricator.wikimedia.org/T166010) that Morne mentioned). In the frontend, WMF has Codex and Vector 2022, which both changed core in a significant way. In the backend, the drop of PHP 7.4 happened after 1.39 and it would be nice to not have to support 7.4.

malberts:
> Is there anything unique for the facet item that can be used as an ID for the <input> element? It is needed to make the label linked to the checkbox/radio.

Each filter list item will have some value, which we could perhaps make CSS-safe.
But what about just using numeric ids (sequential or random)? `facet-1-item-1`, or a GUID. We're not actually submitting a form where we need to care about ids/names.

> In the previous implementation, clicking the label will open a new URL. Do we still need to keep that behavior?

That's essentially what Amazon/Ebay/[zbMath](https://zbmath.org) do. There might or might not be some JS involved in triggering the link, but all those list items have links. Whether we move link generation to the frontend, or directly manipulate the location in the frontend, I think the final behavior is still that clicking a filter list item should reload the page, unless we go with completely replacing the search page with Vue.

JeroenDeDauw:
> Is there anything unique for the facet item that can be used as an ID for the <input> element? It is needed to make the label linked to the checkbox/radio.

A hash of the property ID (associated with the facet) and the value shown might work. Theoretically, you could have multiple facets for the same property. So it is not completely safe. Could you use a randomly generated ID? Or perhaps a hash of the "facet number" (compute that) and the value.

> In the previous implementation, clicking the label will open a new URL. Do we still need to keep that behavior?

There is no requirement for that behavior. We did that because on the standard search page we need the reload.

---

@malberts beat me by 5 seconds :)

PR #61, opened 2025-01-03
Use functional style in translatePropertyStatements

JeroenDeDauw:
@alistair3149 I wonder what you think of this code style. Which version do you prefer

### DIFF
--- a/src/Application/StatementListTranslator.php
+++ b/src/Application/StatementListTranslator.php
@@ -6,6 +6,7 @@ namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
 
 use Wikibase\DataModel\Entity\ItemId;
 use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Statement\Statement;
 use Wikibase\DataModel\Statement\StatementList;
 
 class StatementListTranslator {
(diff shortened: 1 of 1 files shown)


alistair3149:
> I wonder what you think of this code style. Which version do you prefer

I'm indifferent to both. The newer style is cleaner code-wise but less readable, though the pattern is quite common so folks should be able to tell it right away.

malberts:
Sometimes it feels like the for loop reads nicer than _nested_ functional code. In the for loop getting the value and including/excluding happens in the same iteration, in sequence. Whereas with the nested functional code two separate actions are happening.

JeroenDeDauw:
The for loop results in code you can read from top to bottom, but more state you have to keep in mind. The functional approach reads from the inside (nested) to the outside. If you are not used to the later, it is less readable. This is a general functional style tradeoff. So team familiarity is an input to figuring out the optimal choice.

> two separate actions are happening

Do you mean the two arguments to the `array_map`?

malberts:
> Do you mean the two arguments to the array_map?

No, the mapping and filtering are "separate" steps.

In this case it's probably not a big deal to go with the functional syntax. Like I mentioned in https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/51#discussion_r1901990169, I think the original loop had more code, but it ended up with much less, which works fine in functional style now.

In a case like https://github.com/ProfessionalWiki/WikibaseFacetedSearch/blob/6bdcf26fadb7b7e949c88a1609e0823a4eac715c/src/Application/StatementListTranslator.php#L46-L51
the functional style felt nicer from the start. It's still nested, but it feels more like one step.

I don't really know at what point nested functional style becomes too unreadable just for the sake of not having a loop.

JeroenDeDauw:
Right. I was not so happy with the array_filter. Keeping that together with the statementToSearchData is nicer.

PR #67, opened 2025-01-03
Add QueryStringParser

JeroenDeDauw:
Started with an AI-generated foundation and made many changes, both small and large

I could not get the quote handling the AI put in to work correctly so skipped that for now. I also did not yet check what the quote support requirements are, if any.

### DIFF
--- a/src/Application/Query.php
+++ b/src/Application/Query.php
@@ -1,0 +1,32 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
+
+use Wikibase\DataModel\Entity\PropertyId;
+
+final class Query {
+
+	public function __construct(
+		public readonly PropertyConstraintsList $constraints,
+		private readonly string $freeText = ''
+	) {
+	}
+
+	public function getConstraintsForProperty( PropertyId $propertyId ): ?PropertyConstraints {
+		return $this->constraints->getConstraintsForProperty( $propertyId );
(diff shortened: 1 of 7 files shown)


src/Application/Query.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
+
+use Wikibase\DataModel\Entity\PropertyId;
+
+final class Query {
```
malberts, line 9: I assume all these `final`s were added by AI? Or are we explicitly preventing subclassing? (I'm not used to seeing this in our other code, even when classes are not really meant to be subclassed).

JeroenDeDauw: Ah yes, indeed. I did not even notice those.

JeroenDeDauw: _(Jeroen-o1 thought about the final keyword for 5 minutes)_

For our projects there seems little difference in having final slapped on most things. We rarely use protected visibility and almost never use inheritance for code reuse.

While my inner programmer wants to spend even more brain cycles on figuring out what the best policy is. However, productivity Jeroen is saying this is a waste of time. So shrug/whatever.

malberts: Indeed doesn't really matter here. My feeling is to not add it, as we've done in other projects, unless there's a very explicit reason to do so (which I don't think we've ever had, hence why I noticed it).

PR #68, opened 2025-01-03
Drop MW 39 and 41 tests

JeroenDeDauw:


alistair3149:
Should we test for PHP 8.4 as well since it is being supported by master/MW 1.44?

alistair3149 (approved):

PR #72, opened 2025-01-04
Replace deprecated class aliases

malberts:
Follow-up to #68 

The remaining classes without namespaces seem to not have namespaces in MW 1.43.


malberts:
The test failure is weird. Obviously caused by something in this PR, but the actual failure is not obvious (to me).

JeroenDeDauw:
I see tests failed with 42 but worked on 43, so presumably you broke compat with 42. Dropping those tests in https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/73

alistair3149 (approved):
LGTM

Issue #76, opened 2025-01-05
Facets UI: Make facets collapsible


JeroenDeDauw:
Make facet collapsible (Maybe use the [Accordion](https://doc.wikimedia.org/codex/v1.14.0/components/demos/accordion.html#css-only-version) component)

![image](https://github.com/user-attachments/assets/d90d54a6-8f6f-4018-a1f7-5ef3996533ea)
![image](https://github.com/user-attachments/assets/03041738-8009-4857-8df4-71bdd1d6cd03)

PR #81, opened 2025-01-06
Make facet collapsible

alistair3149:
Key changes:
- Use Codex [Accordion](https://doc.wikimedia.org/codex/v1.14.0/components/demos/accordion.htm) component to make facet collapsible (#76)
- Align property count to the end of the facet label, like in the mockup
- Remove border of the facets container to allow more space to be utilized, like the mockup
- Add `expanded` boolean value to the return values of `facetsToViewModel`, indicating the initial state of the accordion component

image


JeroenDeDauw (approved):
Looks good, though note that we decided to not have boolean type facets and instead go with the "No value" or "Any value" options in facets that are configured to include those.

Issue #11, opened 2024-11-28
Facets UI: Apply filter in search


JeroenDeDauw:
Probably depends on
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/31

AC:
* Search query is adjusted by interacting with a facet
* The new search query is submitted so the page reloads with new results

Part of
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/8

PR #82, opened 2025-01-06
Make Facet UI checkbox functional

alistair3149:
Part of #11

Key changes:
- Parse current search query from URL in `FacetUiBuilder`
- Define the selected state of the Facet Checkbox from current URL query
- Clicking on a Facet checkbox will open the search page with the correct filter query to the current URL
- Add data attributes for property and item ID to facet and checkboxes
- Add `SearchUrlBuilder` class to build search URL for facet items

To-do:
- Persist collapsible state across navigation


src/Presentation/FacetUiBuilder.php
```diff
@@ -39,40 +44,46 @@ public function __construct(
 	// TODO: parameter: selected values (from QueryStringParser https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/31)
 	public function createHtml( ItemId $itemType ): string {
 		$this->config->getFacetConfigForInstanceType( $itemType );
-		$this->query = $this->getSearchQueryFromUrl();
+
+		$this->query = ( new QueryStringParser() )->parse( $this->getSearchQueryFromUrl()['search'] );
+		$this->constraints = $this->query->getConstraintsPerProperty();
```
JeroenDeDauw, line 52: $this->query is unused in the class, it could be a local variable.

You could compute `$this->constraints` in its own method. Then in this one you would just have

`$this->constraints = $this->getConstraintsPerProperty()`

... there are other issues, though. Unsure if when resolving all you would have such code.

src/Presentation/FacetUiBuilder.php
```diff
@@ -39,40 +44,46 @@ public function __construct(
 	// TODO: parameter: selected values (from QueryStringParser https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/31)
 	public function createHtml( ItemId $itemType ): string {
 		$this->config->getFacetConfigForInstanceType( $itemType );
-		$this->query = $this->getSearchQueryFromUrl();
+
+		$this->query = ( new QueryStringParser() )->parse( $this->getSearchQueryFromUrl()['search'] );
```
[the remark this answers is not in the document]
JeroenDeDauw: General rule: do not create new instances of services inside of other services. Object graph construction is a dedicated responsibility, and we take care of it in the top-level factory that is `WikibaseFacetedSearchExtension`

src/Presentation/FacetUiBuilder.php
```diff
@@ -150,4 +183,29 @@ private function getItemsHtml( array $items, string $type, string $facetName ):
...
+
+		if ( $selected === true ) {
+			$query['search'] = str_replace( $facetQuery, '', $query['search'] );
+		} else {
+			$query['search'] .= $facetQuery;
+		}
+
+		$urlParts = $this->urlParts;
+		$urlParts['query'] = wfArrayToCgi( $query );
+		return UrlUtils::assemble( $urlParts );
```
JeroenDeDauw, line 208: I see several global method calls, such as 2x `wfCgiToArray` and `UrlUtils::assemble`. Luckily they are all leaves in the call graph and do not touch any state (which would be global). They are defined by the framework (MediaWiki), which is a minus. Arn't there native PHP functions that can be used instead, or OO services provided by MediaWiki (I guess not for the later).

All this URL building could go into its own service. This will be easier to test, and for logic like this we do want full unit test coverage. This could be something like a `FacetItemUrlBuilder` with a `buildFacetItemUrl( string $currentUrl, PropertyId $propertyId )` method.

Except that the current parameters seem incorrect because you based your code on that is in `getExampleListItems`. There is not always an item. List facets have values. These might be string. Or they could be Item Ids, in which case we need to show the label of the referenced Item (i.e. Alice) but use the Item Id (i.e. Q42) in the search string and URL.

So perhaps a SearchPageUrlBuilder with `buildUrl( string $currentUrl, PropertyId $propertyId, string $facetQuery )` where $facetQuery contains whatever the facet generates, so values like `haswbfacet:P1<2020 haswbfacet:P1>2010`. This approach keeps the building of `$facetQuery` separate, so we would need to do this elsewhere. Here we would need to support different facet types, which suggests further classes, but I cannot say that for sure without more research.

src/Presentation/FacetUiBuilder.php
```diff
@@ -17,9 +20,18 @@ class FacetUiBuilder {
...
+
+	/** @var array<string, PropertyConstraints> */
+	private array $constraints = [];
+
 	public function __construct(
 		private readonly TemplateParser $parser,
-		private readonly Config $config // TODO: use
+		private readonly QueryStringParser $queryStringParser,
+		private readonly Config $config, // TODO: use
+		private readonly string $url,
```
JeroenDeDauw, line 33: I strongly suspect that $url should be a parameter in `createHtml`

src/Presentation/FacetUiBuilder.php
JeroenDeDauw: Could could start with implementing `SearchUrlBuilder` and adding unit tests for each case you support. Since that class would has no collaborators, this should be straightforward.

JeroenDeDauw: That would move a bunch of logic out of this PR. You could put some hardcoded or nonsense links in this PR. Perhaps the current URL. Then we already have the other UI changes on master.

malberts: > The query string and URL are currently using only the property ID (e.g. P100) and item ID (e.g. Q100). Do we need to support using the labels as well?

No, if a statement refers to an Item, it will always use the ID (Q100). However, a statement can be a string ([docs](https://www.wikidata.org/wiki/Help:Data_type#String-based_data_types)), in which case the value will be the string value (`foo`).

src/Application/SearchUrlBuilder.php
```diff
@@ -0,0 +1,63 @@
...
+	/**
+	 * @return array<string, string>
+	 */
+	public function getUrlParts(): array {
+		return $this->urlParts;
+	}
+
+	public function setUrlQuery(): void {
+		$this->urlQuery = wfCgiToArray( (string)$this->urlParts['query'] );
+	}
```
malberts, line 35: Since this is a setter, I was expecting a parameter.
It also looks like this method is always called after `setUrlParts`, so perhaps there needs to be a single public method like `setUrl( string $url )` which implements or calls both.

src/Presentation/FacetUiBuilder.php
alistair3149: > No, if a statement refers to an Item, it will always use the ID (Q100). However, a statement can be a string ([docs](https://www.wikidata.org/wiki/Help:Data_type#String-based_data_types)), in which case the value will be the string value (foo).

We should probably revisit this in a follow-up patch since this is more about how we parse facet values into strings.

Currently this patch append/remove `haswbfacet:A=B` to the search query when a facet checkbox is clicked. `A` is a string defined under the `propertyId` key in `facetsToViewModel()`, whereas `B` is a string defined under the `itemId` key in `getExampleListItems()`.

PR #85, opened 2025-01-07
Set up Stylelint and ESLint

### LINKED ISSUE #84: CI: Set up CSS/LESS and JS linting
- Include `stylelint-config-wikimedia` as `devDependencies` in `package.json`
- Include `eslint-config-wikimedia` as `devDependencies` in `package.json`
- Set up lint scripts in `package.json`
- Add lint scripts to `ci.yml`

alistair3149:
Key changes:
- Add WMF configurations for Stylelint and ESLint
- Update linting step in Github Actions to include stylelint and eslint
- Fix existing linting issues in styles and JS raised by the linters

Closes #84


.github/workflows/ci.yml
```diff
@@ -225,5 +225,15 @@ jobs:
...
+        run: npm install
+
+      - name: Lint i18n
+        run: npm run lint:i18n
+
+      - name: Lint styles
+        run: npm run lint:styles
+
+      - name: Lint JS
+        run: npm run lint:js
```
JeroenDeDauw, line 239: I wonder if it is better to have a task for each of those linting checks

malberts: In the past we also had all of these in one CI task, except we used the old Grunt config where there was a single `npm run test`, which forced it into a single task.

The main pain point is `npm install` being slow (in theory). But it might not necessarily be an issue here.

PR #89, opened 2025-01-10
Add placeholder And/Or toggle to facet UI

alistair3149:
Key changes
- Add And/Or toggle to list facets. Each button is set up as an `<a>` element so that it can be used to trigger a page load
- Move Facet out of `Facets.mustache` into a new template `Facet.mustache` for cleaner organization
- Use Codex built-in `cdx-button--fake-button` classes directly to style Codex buttons that are not `<button>`

image

### DIFF
--- a/resources/ext.wikibase.facetedsearch.less
+++ b/resources/ext.wikibase.facetedsearch.less
@@ -15,6 +15,16 @@
 		}
 	}
 
+	&__cdx-button-link {
+		color: @color-base;
+
+		&:hover,
+		&:active,
+		&:visited {
+			color: @color-base;
+		}
+	}
+
 	// Use this class when we are not using a <button> element for cdx-button
 	// Because cdx-button styles rely on the :enabled and :disabled pseudo-classes,
 	// we have to insert the background, hover, and active states manually
@@ -82,6 +92,19 @@
 					border-bottom: 0;
 				}
 			}
+		}
+	}
+
+	&__facet-toggle {
+		margin-bottom: @spacing-100;
+		display: grid;
+		grid-template-columns: 1fr 1fr;
+		border-radius: @border-radius-base;
+		overflow: hidden;
+
+		.cdx-button {
+			border-radius: 0;
+			width: 100%;
 		}
 	}
 
--- a/resources/ext.wikibase.facetedsearch.less
+++ b/resources/ext.wikibase.facetedsearch.less
@@ -5,7 +5,7 @@
 
 	&__icon {
 		// Needed the specificity to override cdx-mixin-css-icon mixin
-		&.wikibase-faceted-search__icon {
+		&.wikibase-faceted-search__icon.wikibase-faceted-search__icon.wikibase-faceted-search__icon {
 			// Icon should always stay in sync with the current text color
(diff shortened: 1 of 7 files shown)


resources/ext.wikibase.facetedsearch.less
```diff
@@ -5,7 +5,7 @@
 
 	&__icon {
 		// Needed the specificity to override cdx-mixin-css-icon mixin
-		&.wikibase-faceted-search__icon {
+		&.wikibase-faceted-search__icon.wikibase-faceted-search__icon.wikibase-faceted-search__icon {
```
malberts, line 8: Is this repetition intentional?

alistair3149: Yes it has to have enough specificity override the different styles from Codex.
There are many selectors in Codex so instead of keeping track of each one, we apply enough specificity using our own classes.

We could use `!important` but that is the nuclear option.

malberts: Does repeating the class 3 times like that make it more specific? Isn't that the same as having:
`<div class="wikibase-faceted-search__icon wikibase-faceted-search__icon wikibase-faceted-search__icon">...</div>`.

alistair3149: Yes it actually does increase the specificity. For example:
- `.class` has a specificity of 0-0-1 (1 point)
- `.class.class` has a specificity of 0-0-2 (2 points)
- `.class.class.class` has a specificity of 0-0-3 (3 points)

Setting `.wikibase-faceted-search__icon` to `currentColor` will make sure that we link the icon color to the current text color. Or if we need to redefine the color, we can just use the `color` property.

malberts: Ah okay. I don't think I've ever seen class repetition being used like that, since there would usually be some other selector before going for `!important`.

src/Presentation/FacetUiBuilder.php
```diff
@@ -54,23 +56,26 @@ private function facetsToViewModel( /* TODO: parameters */ ): array {
 				'values-html' => $this->getItemsHtml(
 					$this->getExampleListItems(), FacetType::LIST->value, 'Author'
 				),
-				'expanded' => true
+				'expanded' => true,
+				'hasToggle' => true
```
malberts, line 60: Not a technical issue, but `hasToggle` and `FacetToggle` seem a bit generic. Although we don't have any other toggles at the moment.

resources/ext.wikibase.facetedsearch.less
alistair3149: Yeah usually it is the case, we can revisit the selector in the future.
The complication is that Codex selectors have a lot of specificity and variants, and having to support multiple Codex version make it difficult to maintain.

src/Presentation/FacetUiBuilder.php
alistair3149: I agree.
The attribute is currently a placeholder and should be shortly swap out since we only use it for list facet.
The value/config should be derived from the facet type instead, when we have more solid spec on the facet config.

PR #93, opened 2025-01-11
Refactor away facetTemplates field

JeroenDeDauw:


src/Presentation/FacetUiBuilder.php
```diff
@@ -148,31 +145,29 @@ private function getExampleRangeItems(): array {
 	/**
 	 * @param array<array<string, mixed>> $items
 	 */
-	private function getItemsHtml( array $items, string $type, string $propertyId ): string {
```
JeroenDeDauw, line 151: @alistair3149 we have the Enum so we can use it as a type. We only transition to the string value when calling code we do not control or where the PHP type is not possible, such as when handing over to the template engine.

src/Presentation/FacetUiBuilder.php
```diff
@@ -181,6 +176,13 @@ private function getItemsHtml( array $items, string $type, string $propertyId ):
 		return $html;
 	}
 
+	private function facetTypeToTemplateName( FacetType $type ): string {
+		return match ( $type ) {
+			FacetType::LIST => 'FacetItemCheckbox',
+			FacetType::RANGE => 'FacetItemRange',
+		};
+	}
```
JeroenDeDauw, line 184: @alistair3149 this is a more functional approach. No field and no $template variable

Issue #40, opened 2024-12-15
Show instance type choosing facet when there is no filter on the type yet


JeroenDeDauw:
Show a list of item types (ie Documents, Authors, Software, Formulae). 

When a item type is picked, add the filter for that to the search, ie `haswbstatement:P1460=Q5976445` for Authors. Then only Authors are shown, and we can show the facets for Auhtors.

This list of item types is not a regular facet. It should always be shown, while the normal facets are linked to an item type, and only show up once the results are filtered for their item type. Once an item type is shown we might wish to condense the display of the item type selector, since changing that type after initial selection is presumably rare, especially if the user used any of the normal facets.

Equivalent to the tabs on zbmath
![image](https://github.com/user-attachments/assets/d10989ad-b8c4-4acc-92f2-7377bb90bb8f)

Somewhat depends on
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/31
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/80 (provides "label" in the config, for display in our list)

Issue #23, opened 2024-11-29
Create PHP service that returns facet values and counts


JeroenDeDauw:
AC:
* support for key value types: item Ids, strings, ??? 
* counts for each value, limited based on already selected filters
 
## Example

* Bob (5)
* Alice (3)
* NyanCat (1)

Actually: we might want to avoid item labels in this service, and return like the below. Another service could do the item label fetching, and this could be on the frontend, using progressive enhancement.

* Q123 (5)
* Q456 (3)
* Q789 (1)

---

ValueCounter added in https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/95, to be turned into interface with elastic and test implementations

Relevant conversation about return types: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/95#discussion_r1913538756

JeroenDeDauw:
## How to run queries via CLI

```
curl -X GET "localhost:9200/_search?pretty" -H "Content-Type: application/json" -d'
{
  "query": {
    "bool": {
      "must": [
        {
          "match": {
            "statement_keywords": "P1460=Q5976445"
          }
        }
      ]
    }
  }
}'
```

Run from a `query.json` file: `curl -X GET "localhost:9200/_search?pretty" -H "Content-Type: application/json" -d @query.json`

## Aggregation progress

This works, but there has to be a better way

```json
{
  "size": 0,
  "query": {
    "match": {
      "statement_keywords": "P1460=Q5976449"
    }
  },
  "aggs": {
    "volumes": {
      "terms": {
        "script": {
          "source": "def volumes = []; if (params['_source'].statement_keywords != null) { for (def statement : params['_source'].statement_keywords) { if (statement.startsWith('P16=')) { volumes.add(statement.substring(4)); } } } return volumes;",
          "lang": "painless"
        },
        "size": 10,
        "order": { "_count": "desc" }
      }
    }
  }
}
```

Output:

```json
{
  "took" : 19,
  "timed_out" : false,
  "_shards" : {
    "total" : 3,
    "successful" : 3,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 50,
      "relation" : "eq"
    },
    "max_score" : null,
    "hits" : [ ]
  },
  "aggregations" : {
    "volumes" : {
      "doc_count_error_upper_bound" : 1,
      "sum_other_doc_count" : 135,
      "buckets" : [
        {
[...]

PR #95, opened 2025-01-11
Redesign UI building code

JeroenDeDauw:
I redid the structure from scratch, and copied over various bits from the previous implementation. Many local TODOs left. Some regression in function (ie links) and some gain (ie real config used).


JeroenDeDauw:
For review, please pay particular attention to the mustche templates. I tried to keep them as close to the originals as possible but there might be accidental changes. Remember I can be quite blind when it comes to small details. For some reason the counts no longer show up, I could not find the issue within 5 minutes. The template is getting the expected numbers.

src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -0,0 +1,71 @@
...
+		$selectedValues = $combineWithAnd ? $state->getAndSelectedValues() : $state->getOrSelectedValues();
+
+		$checkboxes = [];
+
+		foreach ( $this->getValuesAndCounts( $config ) as $i => $valueCount ) {
+			$checkboxes[] = [
+				'label' => $valueCount->value,
+				'count' => $valueCount->count, // FIXME: count is now showing in the UI for some reason
+				'checked' => in_array( $valueCount->value, $selectedValues ), // TODO: test with multiple types of values
+				'link' => 'TODO', // TODO: link values (probably need new collaborator for URL stuff, see SearchUrlBuilder)
```
JeroenDeDauw, line 54: To build the links here, we need the $currentUrl. We could add such a parameter to `FacetHtmlBuilder::buildHtml`, but let's explore our options, as it's not obvious to me that is the best approach.

To me, the simplest approach is to not build links here at all, and instead have some JS on the FE. @alistair3149 went with the URL approach instead of modifying the search query in the form so we can include information about the collapsed state of facets and similar UI info. That info is not available on the backend, so that approach is not compatible with the current one of building the links in PHP.

There also was some talk about potentially allowing multiple value selection before reload and of a "deselect all". These are not important ATM, but also benefit from the client side JS approach.

For the MVP we do need range facets. In there we can't have links. Presumably we need JS there. 

TL;DR: looks to me like we should not build links and instead have data-attributes for JS to act on

Concerns with dropping the links in favor of JS?

src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -0,0 +1,71 @@
...
+
+				// TODO: can't we escape this in the template?
+				'id' => Sanitizer::escapeIdForAttribute( htmlspecialchars( $state->propertyId->getSerialization() . "-$i" ) ),
```
malberts, line 57: `{{id}}` in Mustache is for HTML escaping. CSS ID escaping is different (e.g. whitespace is not allowed).

src/Presentation/UiBuilder.php
```diff
@@ -0,0 +1,74 @@
...
+
+		foreach ( $this->config->getFacetConfigForInstanceType( $itemType ) as $facetConfig ) {
+			$facets[] = $this->buildFacetViewModel(
+				$facetConfig,
+				$query->getConstraintsForProperty( $facetConfig->propertyId ) ?? new PropertyConstraints( $facetConfig->propertyId )
+			);
+		}
+
+		return $facets;
+	}
```
malberts, line 62: Should we use `array_map` here?

templates/ListFacet.mustache
```diff
@@ -0,0 +1,41 @@
...
+					type="checkbox"
+					{{#checked}}checked{{/checked}}
+					{{! Prevent input from being interacted by keyboard}}
+					tabindex="-1"
+				/>
+				<span class="cdx-checkbox__icon"></span>
+				<div class="cdx-checkbox__label cdx-label">
+					<label class="cdx-label__label" for="{{id}}">
+						{{! Change to <div> tag instead of <span> tag for valid HTML }}
+						<span class="cdx-label__label__text">{{>FacetItemLabel}}</span>
```
malberts, line 34: What flagged this as invalid HTML? It's currently `<span>` with more `<span>`s inside, which is valid.
Strictly speaking, having `<div>` inside `<label>` is invalid HTML (see "Permitted content" https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#technical_summary)

src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -0,0 +1,71 @@
...
+	/**
+	 * TODO: add integration smoke test
+	 * TODO: add unit tests for the logic. Likley requires some refactoring, ie making the view model accessible
+	 */
+	public function buildHtml( FacetConfig $config, PropertyConstraints $state ): string {
+		return $this->parser->processTemplate(
+			'ListFacet',
+			[
+				'hasToggle' => true, // TODO: use config allowCombineWithChoice and defaultCombineWith
+				'checkboxes' => $this->buildCheckboxesViewModel( $config, $state ),
```
malberts, line 34: Since `DelegatingFacetHtmlBuilder` works on `FacetType` only, are we going to have a different `FacetType` and `FacetHtmlBuilder` for single value radio buttons? Or is that not really a use case any more?
It seems like only the instance type facet might have a single value, but whether we implement that as another facet is not clear (https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/40).

templates/ListFacet.mustache
malberts, line 34: > For some reason the counts no longer show up,

In `FacetItemLabel.mustache` we can remove the `{{#count}}` condition.

Screenshot_20250113_130941


I don't think we need that any more, since any list facet value that we send to the template must have a count (`ValueCount $count` is not null). If it has 0 count then we shouldn't display the facet value (and even if we had to, the count is still a number).

malberts (approved):

src/Presentation/ListFacetHtmlBuilder.php
JeroenDeDauw: I would not account for the instance type selection in the facet system.

src/Presentation/ListFacetHtmlBuilder.php
alistair3149: Yes since we don't need to support browsers without JS, we can build the interaction through JS and pass the data through data attributes.

templates/ListFacet.mustache
alistair3149: Yes you are correct.
It was a leftover comment from an older version of the `FacetItemLabel` that uses a different HTML.
I will remove it in the next PR

src/Application/PropertyConstraints.php
```diff
@@ -45,13 +45,15 @@ public function hasNoValue(): bool {
 	}
 
 	/**
+	 * TODO: is this always string?
```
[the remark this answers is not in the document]
JeroenDeDauw: image

In this code we will need to make the types match, $selectedValues comes from the QueryStringParser, so the code we commented on in this PR, and $valueCount->value comes from the to-be-implemented ValueCounter (https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/23). 

So will indeed need to use the confus to changes the types of the parsed values, unless we also get strings from the elastic implementation of ValueCounter. Unclear if we should put config based conversation in the query string parser or its caller. Either way, no change needed now.

PR #100, opened 2025-01-13
Implement apply button for range facet

alistair3149:


src/Presentation/RangeFacetHtmlBuilder.php
```diff
@@ -22,13 +22,18 @@ public function __construct(
...
 			'RangeFacet',
 			[
 				'msg-min' => wfMessage( 'wikibase-faceted-search-facet-range-min' )->text(),
 				'msg-max' => wfMessage( 'wikibase-faceted-search-facet-range-max' )->text(),
-				'current-min' => $state->getInclusiveMinimum(),
-				'current-max' => $state->getInclusiveMaximum(),
+				'msg-apply' => wfMessage( 'wikibase-faceted-search-facet-apply' )->text(),
+				'current-min' => $currentMin,
+				'current-max' => $currentMax,
+				'disabled' => $currentMin === null || $currentMax === null
```
malberts, line 36: We allow open-ended ranges, so the button should not be disabled if any of the values are empty.

malberts: We could disable the button by default and only enable it once any value changes. But that would probably be frontend-only and would not require passing additional data to the template.

malberts (changes requested):

src/Presentation/RangeFacetHtmlBuilder.php
alistair3149: In that case then I will just remove the `disabled` parameter and make it the default, and have the frontend remove the `disabled` attribute

src/Presentation/RangeFacetHtmlBuilder.php
```diff
@@ -22,13 +22,17 @@ public function __construct(
...
 		return $this->parser->processTemplate(
 			'RangeFacet',
 			[
 				'msg-min' => wfMessage( 'wikibase-faceted-search-facet-range-min' )->text(),
 				'msg-max' => wfMessage( 'wikibase-faceted-search-facet-range-max' )->text(),
-				'current-min' => $state->getInclusiveMinimum(),
-				'current-max' => $state->getInclusiveMaximum(),
+				'msg-apply' => wfMessage( 'wikibase-faceted-search-facet-apply' )->text(),
+				'current-min' => $currentMin,
+				'current-max' => $currentMax
```
JeroenDeDauw, line 35: Why did you introduce variables here? 

I think the old form is better/simpler, though it does not matter much as this is all local

src/Presentation/RangeFacetHtmlBuilder.php
JeroenDeDauw: Looks like you want to do that in a follow up?

alistair3149: > Looks like you want to do that in a follow up?

Yes that part will be implemented when we make the range facet functional on the frontend

src/Presentation/RangeFacetHtmlBuilder.php
alistair3149: It was a leftover from [bcc1cfc](https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/100/commits/bcc1cfcaa470b91491fbbea6aaa6987ba2d1baf8). I have created a follow-up PR: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/103

PR #101, opened 2025-01-13
Define toggle states in ListFacetHtmlBuilder

alistair3149:
Key changes:
- Implement different states of the toggle
- Move toggle template data out of buildHtml into a new function
- Add `data-value-id` attribute to facet checkbox items

### DIFF
--- a/src/Presentation/ListFacetHtmlBuilder.php
+++ b/src/Presentation/ListFacetHtmlBuilder.php
@@ -50,11 +50,15 @@ class ListFacetHtmlBuilder implements FacetHtmlBuilder {
 			'and' => [
 				'label' => wfMessage( 'wikibase-faceted-search-and' )->text(),
 				'selected' => $combineWithAnd,
+				// TODO: Remove the ignore when dynamic value for $combineWithAnd is implemented
+				// @phpstan-ignore ternary.alwaysTrue
 				'disabled' => $combineWithAnd === true ? false : $disabled
 			],
 			'or' => [
 				'label' => wfMessage( 'wikibase-faceted-search-or' )->text(),
 				'selected' => !$combineWithAnd,
+				// TODO: Remove the ignore when dynamic value for $combineWithAnd is implemented
+				// @phpstan-ignore ternary.alwaysTrue
 				'disabled' => $combineWithAnd === false ? false : $disabled
 			]
 		];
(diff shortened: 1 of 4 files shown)


templates/ListFacet.mustache
```diff
@@ -1,29 +1,44 @@
...
+        class="cdx-checkbox__input"
+        type="checkbox"
+        {{#checked}}checked{{/checked}}
+        {{! Prevent input from being interacted by keyboard}}
+        tabindex="-1"
+      />
+      <span class="cdx-checkbox__icon"></span>
+      <div class="cdx-checkbox__label cdx-label">
+        <label class="cdx-label__label" for="{{id}}">
+          {{! Change to <div> tag instead of <span> tag for valid HTML }}
```
JeroenDeDauw, line 39: I saw this comment before and do not know what I am supposed to take away from it

src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -51,6 +70,7 @@ private function buildCheckboxesViewModel( FacetConfig $config, PropertyConstrai
 				'label' => $valueCount->value,
 				'count' => $valueCount->count, // FIXME: count is now showing in the UI for some reason
 				'checked' => in_array( $valueCount->value, $selectedValues ), // TODO: test with multiple types of values
+				'value-id' => $valueCount->value,
```
JeroenDeDauw, line 73: This does not make sense

I guess you put this here because for values of type `Item ID` we want to show the associated Wikibase Item label instead of the ID itself. In those cases we could theoretically provide the label to the label key here, and the value itself to `value-id`. Values of other types do not have an ID, so it does not work for them.

There also is uncertainty about where we want to get the label for item ids. We could do this on the frontend and get ourselves form progressive enhancement. 

For now, just ingore the item ID complications. What we need in the UI is the value itself, since they is what we need in the query string. We can decide how to show labels instead of IDs later.

alistair3149: The attribute is only intended to be consumed by the JS.

I am not sure what would be the correct term for the value that we use for the `haswbfacet` query, so I put `value-id` for now as it is an identifier for the value. Is there a more suitable term that can be used for the data attribute?

JeroenDeDauw: `value`. I suppose renaming to that would be OK for now, though we might end up only needing `value` and not `label`

templates/ListFacet.mustache
alistair3149: It is caused by the previous merge, I have created a follow-up here: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/104

PR #102, opened 2025-01-13
Allow checkbox to submit search query

alistair3149:
Part of: #11

Key changes:
- Attach an event to facet checkbox that submit the correct search query on click
- Add `ext.wikibase.facetedsearch` ResourceLoader module that contains the JS needed

To-dos:
- Support OR query
- Support inverse query

### DIFF
--- a/resources/ext.wikibase.facetedsearch.js
+++ b/resources/ext.wikibase.facetedsearch.js
@@ -52,14 +52,13 @@ function buildQueryString( oldQuery, facet ) {
 
 	// Add selected facet items to the query string
 	[ ...facet.querySelectorAll( '.wikibase-faceted-search__facet-item' ) ].forEach( ( facetItem ) => {
-		const value = facetItem.dataset.valueId;
 		const checkbox = facetItem.querySelector( '.cdx-checkbox__input' );
-		if ( !value || !checkbox || !checkbox.checked ) {
+		if ( !checkbox || !checkbox.checked || !checkbox.value ) {
 			return;
 		}
 		// TODO: Support other operators
 		// TODO: Support OR values
-		queries.push( `haswbfacet:${ propertyId }=${ value }` );
+		queries.push( `haswbfacet:${ propertyId }=${ checkbox.value }` );
 	} );
 
 	return queries.join( ' ' );
--- a/resources/ext.wikibase.facetedsearch.js
+++ b/resources/ext.wikibase.facetedsearch.js
@@ -4,24 +4,27 @@ let input;
  * Main entry point for the JavaScript code.
  */
 function init() {
-	const checkboxItems = document.querySelectorAll( '.wikibase-faceted-search__facet-item-checkbox' );
-	if ( !checkboxItems ) {
+	const facets = document.querySelector( '.wikibase-faceted-search__facets' );
+	if ( !facets ) {
 		return;
 	}
 
 	input = document.querySelector( '#searchText > input' );
 
-	checkboxItems.forEach( ( item ) => item.addEventListener( 'change', onCheckboxItemChange ) );
+	facets.addEventListener( 'change', onFacetsChange );
 }
 
 /**
- * Handles the change event for a checkbox item in the faceted search.
+ * Handles the change event for any input elements in the facets.
  *
- * @param {Event} event - The change event from the checkbox.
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
Conflicts with master

resources/ext.wikibase.facetedsearch.js
```diff
@@ -0,0 +1,53 @@
...
+	if ( input.value.indexOf( query ) > -1 ) {
+		// If it is, remove it from the query
+		input.value = input.value.replace( query, '' ).trim();
+	} else {
+		// If it's not, append it to the query
+		input.value += ` ${ query }`;
+	}
+
+	// Submit the search form
+	input.form.submit();
```
JeroenDeDauw, line 50: Better to keep this side effect and the logic separate. Ideally we have a pure function that outputs the entire new query string and as input gets ... the current string and the event or part of the event?

Would look roughly like

```js
function onCheckboxItemChange( event ) {
    input.value = buildQueryString( input.value, event.currentTarget );
}

function buildQueryString( oldQuery, checkbox ) {
    // TODO logic. This might be split into further pure functions
    return newQueryString
}
```

Then it becomes easy to write tests for `buildQueryString`

I'd also change the logic a little. Now that we have only one facet per property, a facet for P123 basically owns all `haswbstatement:P123*`. So I would drop all P123 query segments and then add the newly computed query segments for P123. Otherwise you might keep junk in the URL that can cause confusion in cases where it makes the search results deviate from what you would expect purely based on the UI state.

JeroenDeDauw: The logic approach I described is also more robust as it does not assume we rebuild the query string on each checkbox change, which we would not want to do if we have a `select none` or `select all`.

So we would drop all P123 segments (or whatever the property id associated with the facet is), and then, **on the facet level**, build the new segments for P123, holding into account all the checkboxes.

alistair3149: That approach sounds logical and more solid.
I will refactor this PR tomorrow, moving this to draft for now.

resources/ext.wikibase.facetedsearch.js
```diff
@@ -0,0 +1,78 @@
+let input;
```
JeroenDeDauw, line 1: This is a bit generic. Same with `toggle` elsewhere in recently added code. There are also technical terms that say nothing about the domain.

`searchInput` is clearer

Scope also matters. For variable names or constants, the bigger the scope, the bigger the name. So inside of a short function, `input` might be optimal, but for this case we have something used in the entire file.

resources/ext.wikibase.facetedsearch.js
```diff
@@ -0,0 +1,78 @@
...
+		if ( !checkbox || !checkbox.checked || !checkbox.value ) {
+			return;
+		}
+		// TODO: Support other operators
+		// TODO: Support OR values
+		queries.push( `haswbfacet:${ propertyId }=${ checkbox.value }` );
+	} );
+
+	return queries.join( ' ' );
+}
```
JeroenDeDauw, line 65: Some functions can be extracted

If you find yourself writing comments for blocks of code in a function, that is a signal that those blocks maybe should be their own functions. Comments are rarely needed if you do one thing on one level of attraction per function and you name your functions well.

JeroenDeDauw: Btw, function names behave oposite of variables wrt optimal length. Functions in small scopes often have long names.

resources/ext.wikibase.facetedsearch.js
alistair3149: I agree that we should have it more specific.
I'm not sure about `searchInput` because it is associated with search box (`#searchInput`) provided by MediaWiki skins.
What do you think about a term such as `specialSearchInput`?

JeroenDeDauw: I got onto `searchInput` because the selector is `#searchText > input`

I have some aversion to  `specialSearchInput` due to `special`. At least it isn't magic :) 

From the context it probably is clear what "search input" we refer to, so specifying it is the one added by the special page might be overly specific.

_Bikeshedding alert, spending too much time on this_

PR #105, opened 2025-01-14
Add ListFacetHtmlBuilderTest foundation

JeroenDeDauw:
Starting with an integration test. Next ones will be unit tests via buildViewModel

### DIFF
--- a/src/Presentation/ListFacetHtmlBuilder.php
+++ b/src/Presentation/ListFacetHtmlBuilder.php
@@ -22,10 +22,6 @@ class ListFacetHtmlBuilder implements FacetHtmlBuilder {
 	) {
 	}
 
-	/**
-	 * TODO: add integration smoke test
-	 * TODO: add unit tests for the logic
-	 */
 	public function buildHtml( FacetConfig $config, PropertyConstraints $state ): string {
 		return $this->parser->processTemplate(
 			'ListFacet',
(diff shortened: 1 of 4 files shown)


src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -68,7 +71,7 @@ private function buildCheckboxesViewModel( FacetConfig $config, PropertyConstrai
 		foreach ( $this->getValuesAndCounts( $config ) as $i => $valueCount ) {
 			$checkboxes[] = [
 				'label' => $valueCount->value,
-				'count' => $valueCount->count, // FIXME: count is now showing in the UI for some reason
+				'count' => $valueCount->count,
 				'checked' => in_array( $valueCount->value, $selectedValues ), // TODO: test with multiple types of values
 				'value-id' => $valueCount->value,
 				// TODO: can't we escape this in the template?
```
alistair3149, line 77: We can remove this TODO.
Mustache escaping is not sufficient for that since we need to escape whitespace as well.

JeroenDeDauw: Is unrelated to this PR

The value is guaranteed to not contain spaces because we do not put anything in there that can have spaces. It's also guaranteed to "not need" escaping, but for security in layers, all computed values that are output are still escaped.

JeroenDeDauw: So my current susicion is that we can remove the escaping here because mustache is already doing it. Needs verification first tho

alistair3149: Yes Mustache already escapes HTML. If we don't have any whitespace in value then we can just drop Sanitizer and use Mustache directly.

PR #111, opened 2025-01-14
Act on defaultCombineWith and allowCombineWithChoice config

JeroenDeDauw:


src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -33,37 +38,63 @@ public function buildHtml( FacetConfig $config, PropertyConstraints $state ): st
...
+			if ( $state->getAndSelectedValues() !== [] ) {
+				return true;
+			}
+		}
+
+		if ( ( $config->typeSpecificConfig[self::CONFIG_KEY_DEFAULT_COMBINE_WITH] ?? null ) === 'OR' ) {
+			return false;
+		}
+
+		if ( ( $config->typeSpecificConfig[self::CONFIG_KEY_DEFAULT_COMBINE_WITH] ?? null ) === 'AND' ) {
```
JeroenDeDauw, line 66: I originally did not have the brackets around the `foo ?? null`. Another case of "if it is not tested, it is broken" :)

src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -33,37 +38,63 @@ public function buildHtml( FacetConfig $config, PropertyConstraints $state ): st
...
+			if ( $state->getOrSelectedValues() !== [] ) {
+				return false;
+			}
+
+			if ( $state->getAndSelectedValues() !== [] ) {
+				return true;
+			}
+		}
+
+		if ( ( $config->typeSpecificConfig[self::CONFIG_KEY_DEFAULT_COMBINE_WITH] ?? null ) === 'OR' ) {
```
alistair3149, line 62: Should we define `AND` and `OR` as a constant? Or is it okay for this case?

JeroenDeDauw: Unclear what would be gained by using a constant in this case

JeroenDeDauw: Thinking about it a bit: it is rather arbitrary to use constants for the config keys and not for the values. With the small exception of CONFIG_KEY_DEFAULT_COMBINE_WITH being used multiple times. Using constants for the AND and OR values seems marginally better

alistair3149: > Unclear what would be gained by using a constant in this case

The same value was used in multiple places so I'm wondering if it is worth it to define as a constant.

JeroenDeDauw: I see only one usage of AND and OR. Are you including the tests in your count? In tests there is reason not to use production constants to avoid missing typos in them.

tests/Presentation/ListFacetHtmlBuilderTest.php
```diff
@@ -96,19 +97,120 @@ public function testOnlyCheckboxesForValuesInThePropertyConstraintsAreChecked_an
...
+		$viewModel = $this->buildViewModel(
+			constraints: $this->newPropertyConstraints(),
+			typeSpecificConfig: [ 'allowCombineWithChoice' => false, 'defaultCombineWith' => 'OR' ]
+		);
+
+		$this->assertTrue( $viewModel['toggle']['and']['disabled'] );
+		$this->assertFalse( $viewModel['toggle']['or']['disabled'] );
+	}
+
+	public function testOrIsDisabledWhenOrIsSelectedAndChoiceIsDisabled(): void {
```
malberts, line 124: Should this be `WhenAndIsSelected`?

JeroenDeDauw: Nice catch. Fixed on master

PR #113, opened 2025-01-14
Refactor search result title rewrite

### LINKED ISSUE #75: Clean up onShowSearchHitTitle
And the private static functions it uses:
*  getItemId,
* rewriteLinkForNonEntityResult
* termFallbackToTermData

AC:
* Logic in instance scope (unless reason to deviate)
* Logic tested (unless reason to deviate)

### LINKED ISSUE #94: Rewrite Page search result link text
Follow-up to https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/4

Related refactor: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/75

Initially we rewrote only the Wikibase Item search result link. However, after https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/49 we are indexing data on Pages, and therefore getting Page search results. This Page link text needs to be rewritten to the sitelinked Item's label.

image

malberts:
Fixes https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/75 and https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/94

* Removes the old code for rewriting an Item search result to link to the sitelinked page.
* Rewrites Page search result link title to Item's title

Before:
Screenshot_20250117_022837

After:
Screenshot_20250117_022806


src/EntryPoints/WikibaseFacetedSearchHooks.php
```diff
@@ -49,94 +41,18 @@ public static function onShowSearchHitTitle(
...
-		if ( $term ) {
-			return [
-				'value' => $term->getText(),
-				'language' => $term->getActualLanguageCode(),
-			];
-		}
-
-		return null;
+		$titleSnippet = $rewrittenLink->titleSnippet ?? $titleSnippet;
+		$attributes = $rewrittenLink->attributes + $attributes;
```
malberts, line 55: I went with a little data class here, instead of mutating the parameters. However, this means some logic is outside of the new service. Thoughts on this versus just mutating in the method?

JeroenDeDauw: "the new service" = `PageSearchResultRewriter`

"the method" = ?

malberts: `getRewrittenLink()`

malberts:
I decided to greatly simplify this and remove YAGNI stuff. Some of the original code in this PR was based on the old code that tried to rewrite Item links in a way that was consistent with how Item links are built by Wikibase. However, since we really just need to rewrite the text for a Page link, which is not generated by Wikibase, we don't need to maintain consistency with Wikibase link generation.

src/EntryPoints/WikibaseFacetedSearchHooks.php
JeroenDeDauw: Does this question still matter? I cannot find the code in the latest diff

PR #115, opened 2025-01-15
Make range facet functional

alistair3149:
Part of #11

Key changes:
- Make range facet functional and trigger search form
- Switch to `input` event from `change` for Facet so it can listen for text input such as the ones in range facet
- Separate the generation of facet query from `buildQueryString()`

### DIFF
--- a/resources/ext.wikibase.facetedsearch.js
+++ b/resources/ext.wikibase.facetedsearch.js
@@ -21,9 +21,12 @@ function init() {
  */
 function onFacetsInput( event ) {
 	const input = event.target;
-	const facet = input.closest( '.wikibase-faceted-search__facet' );
+	if ( !input ) {
+		return;
+	}
 
-	if ( !facet || !input ) {
+	const facet = input.closest( '.wikibase-faceted-search__facet' );
+	if ( !facet ) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


resources/ext.wikibase.facetedsearch.js
```diff
@@ -57,7 +100,41 @@ function buildQueryString( oldQuery, facet ) {
...
+ */
+function getRangeFacetQueries( minInput, maxInput, propertyId ) {
+	const queries = [];
+	if ( minInput.value !== '' ) {
+		queries.push( `haswbfacet:${ propertyId }>=${ minInput.value }` );
+	}
+	if ( maxInput.value !== '' ) {
+		queries.push( `haswbfacet:${ propertyId }<=${ maxInput.value }` );
+	}
+	return queries;
```
JeroenDeDauw, line 123: I'd rename to `getRangeFacetQuerieSegments`

and `return segments`

resources/ext.wikibase.facetedsearch.js
```diff
@@ -57,7 +100,41 @@ function buildQueryString( oldQuery, facet ) {
...
 	} );
+	return queries;
+}
+
+/**
+ * Constructs an array of range facet queries based on the provided minimum
+ * and maximum input values.
+ *
+ * @param {HTMLInputElement} minInput - The input element for the minimum range value.
+ * @param {HTMLInputElement} maxInput - The input element for the maximum range value.
```
JeroenDeDauw, line 111: Looks like the code only needs the value, not the HTML elements. That will also make it easier to test.

I like that you documented all the types. Dunno about the descriptions tho. There is no requirement to (not) add them. Best to omit if they do not add value (by saying anything not obvious)

Issue #121, opened 2025-01-15
Set up JS unit testing


JeroenDeDauw:
AC:
* Can run locally
* Runs on GitHub Actions

Ideally we keep this consistent with the most recent versions of similar testing we do elsewhere. Idk if there is something applicable, since our recent frontend testing uses other tech stacks

Issue #112, opened 2025-01-14
Add unit tests for JS query string modification


JeroenDeDauw:
Depends on
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/121

Follow up to
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/102

alistair3149:
Is there a preferred library to use for JS unit tests?
If not then I can set up Jest for the repo.

malberts:
Jest is fine, and is also recommended by MW. We used Vitest for other more complicated Typescript code, but here it will likely be overkill.

JeroenDeDauw:
I created a dedicated task for the setup itself
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/121

PR #123, opened 2025-01-15
Set up jest

alistair3149:
Part of #121

Key changes:
- Add `jest` and `jest-environment-jsdom` as dev dependencies for NPM
- Add `jest` to GitHub CI
- Move PHP tests to `tests/phpunit`
- Add rule to prevent `phpcs` from linting non-PHP files

To-dos:
- Write tests for `ext.wikibase.facetedsearch.js` after #115 is merged
- Figure out how to upload coverage from jest


.github/workflows/ci.yml
```diff
@@ -215,8 +215,8 @@ jobs:
 
       - run: vendor/bin/phpcs -p -s
 
-  linting:
-    name: "Linting"
+  npm:
+    name: "NPM: Lint and unit test"
```
JeroenDeDauw, line 219: Putting different linting tasks together is one thing, also including tests is another. I would not combine these. Nice to see the task level what type of failure you are dealing with.

phpcs.xml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0"?>
 <ruleset>
 	<file>src/</file>
-	<file>tests/</file>
+	<file>tests/phpunit</file>
+
+	<arg name="extensions" value="php"/>
```
JeroenDeDauw, line 6: What is this for?

JeroenDeDauw (approved):

phpcs.xml
alistair3149: phpcs was linting the JS files: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/actions/runs/12794862693/job/35670904144

This is added to make sure that phpcs only touches PHP files

JeroenDeDauw (approved):

phpcs.xml
malberts: Is this still necessary if the paths are limited to (PHP) src and tests?

tests/jest/ext.wikibase.facetedsearch.test.js
```diff
@@ -0,0 +1,19 @@
+describe( 'ext.wikibase.facetedsearch.js', () => {
+	// TODO: Add more tests when the JS is more stable
```
malberts, line 2: Would it be able to add a stub test for the actual JS file in the meantime? It could just load the JS and without the necessary HTML `init()` should just return. Just as an example, since it's unclear how the current JS will be tested.

alistair3149: > Would it be able to add a stub test for the actual JS file in the meantime? It could just load the JS and without the necessary HTML `init()` should just return. Just as an example, since it's unclear how the current JS will be tested.
> 
> Otherwise if you have tests almost ready on an incoming PR, then ignore this.

In the current JS, the functions aren't exported so it can't be accessed by the test. I am going to refactor the JS for #112 after we gets the other JS-related tasks in the MVP milestone.

PR #127, opened 2025-01-16
Make instance tabs functional

alistair3149:
#40

Key changes:
- Parse instance from query string, and use it for the selected state for instance tab
- Add basic instance handling to `Query` and `QueryStringParser`
- Make facets view model based on current instance
- Change `getFilteredQueries()` to remove all WBFS query when `propertyId` is not provided
- Sync instance type example values with config in Docker image

To-dos:
- Add tests
- Create a new class for instance
- Get instances view model from config

### DIFF
--- a/src/Application/Query.php
+++ b/src/Application/Query.php
@@ -8,6 +8,9 @@ use Wikibase\DataModel\Entity\PropertyId;
 
 class Query {
 
+	/**
+	 * @param ItemId[] $itemTypes
+	 */
 	public function __construct(
 		public readonly PropertyConstraintsList $constraints,
 		private readonly string $freeText = '',
--- a/src/Application/Query.php
+++ b/src/Application/Query.php
@@ -4,14 +4,10 @@ declare( strict_types = 1 );
 
 namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
 
-use Wikibase\DataModel\Entity\ItemId;
 use Wikibase\DataModel\Entity\PropertyId;
 
 class Query {
 
-	/**
-	 * @param array<string, NumericPropertyId|ItemId>|array $instance
-	 */
 	public function __construct(
 		public readonly PropertyConstraintsList $constraints,
 		private readonly string $freeText = '',
--- a/src/Application/Query.php
+++ b/src/Application/Query.php
@@ -15,7 +15,7 @@ class Query {
 	public function __construct(
 		public readonly PropertyConstraintsList $constraints,
 		private readonly string $freeText = '',
-		private readonly array $instance = []
+		private readonly array $itemTypes = []
 	) {
 	}
 
@@ -34,12 +34,8 @@ class Query {
 		return $this->freeText;
 	}
 
-	public function getInstancePropertyId(): ?PropertyId {
-		return $this->instance['propertyId'] ?? null;
-	}
-
-	public function getInstanceItemId(): ?ItemId {
-		return $this->instance['itemId'] ?? null;
+	public function getInstanceItemTypes(): array {
+		return $this->itemTypes;
 	}
 
 }
--- a/src/Application/Query.php
+++ b/src/Application/Query.php
@@ -4,13 +4,15 @@ declare( strict_types = 1 );
 
 namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
 
+use Wikibase\DataModel\Entity\ItemId;
 use Wikibase\DataModel\Entity\PropertyId;
 
 class Query {
 
 	public function __construct(
 		public readonly PropertyConstraintsList $constraints,
-		private readonly string $freeText = ''
+		private readonly string $freeText = '',
+		private readonly array $instance = []
 	) {
 	}
 
@@ -27,6 +29,14 @@ class Query {
 
 	public function getFreeText(): string {
 		return $this->freeText;
+	}
+
+	public function getInstancePropertyId(): ?PropertyId {
+		return $this->instance['propertyId'] ?? null;
+	}
+
+	public function getInstanceItemId(): ?ItemId {
+		return $this->instance['itemId'] ?? null;
 	}
 
 }
--- a/src/Application/Query.php
+++ b/src/Application/Query.php
@@ -4,6 +4,7 @@ declare( strict_types = 1 );
 
 namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
 
+use Wikibase\DataModel\Entity\ItemId;
 use Wikibase\DataModel\Entity\PropertyId;
 
 class Query {
--- a/src/Application/Query.php
+++ b/src/Application/Query.php
@@ -9,6 +9,9 @@ use Wikibase\DataModel\Entity\PropertyId;
 
 class Query {
 
+	/**
+	 * @param array<string, NumericPropertyId|ItemId>|array $instance
+	 */
 	public function __construct(
 		public readonly PropertyConstraintsList $constraints,
 		private readonly string $freeText = '',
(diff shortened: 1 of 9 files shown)


src/Application/QueryStringParser.php
```diff
@@ -38,11 +43,33 @@ private function splitQueryString( string $queryString ): array {
...
+	 */
+	private function handleInstancePart( string $part, array &$instance ): array {
+		$part = substr( $part, strlen( 'haswbstatement:' ) );
+		[ $propertyIdString, $itemIdString ] = explode( '=', $part, 2 );
+
+		if ( $propertyIdString === '' || $itemIdString === '' ) {
+			return [];
+		}
+
+		$instance['propertyId'] = new NumericPropertyId( $propertyIdString );
```
alistair3149, line 67: @JeroenDeDauw I am not sure what type should whether we should keep `propertyId` as the type or we should just use string, since it is only used by `UiBuilder`.

src/Application/QueryStringParser.php
```diff
@@ -38,11 +43,33 @@ private function splitQueryString( string $queryString ): array {
...
+		$instance['itemId'] = new ItemId( $itemIdString ); // TODO: Support non-item id values
```
JeroenDeDauw, line 68: This will always be an item id, so the todo is not needed

src/Presentation/UiBuilder.php
```diff
@@ -26,19 +27,27 @@ public function __construct(
 	 * TODO: add integration tests
 	 */
 	public function createHtml( string $searchQuery ): string {
+		$query = $this->parseQuery( $searchQuery );
+		$itemType = $query->getInstanceItemId();
```
JeroenDeDauw, line 31: We have a name mismatch here. Existing code uses `ItemType`, while the new code uses `InstanceItemId` to refer to the same concept.

src/Application/Query.php
```diff
@@ -29,4 +34,12 @@ public function getFreeText(): string {
 		return $this->freeText;
 	}
 
+	public function getInstancePropertyId(): ?PropertyId {
+		return $this->instance['propertyId'] ?? null;
```
JeroenDeDauw, line 38: This information is at Config->instanceOfId, we do not need to parse it out of the string

src/Presentation/UiBuilder.php
```diff
@@ -52,32 +61,46 @@ private function parseQuery( string $searchQuery ): Query {
 	/**
 	 * @return array<array<string, string>>
 	 */
-	private function buildInstancesViewModel(): array {
-		// TODO: Get instances from config
-		// TODO: Get currently selected instance from query
-		return [
+	private function buildInstancesViewModel( ?PropertyId $instancePropertyId, ?ItemId $instanceItemId ): array {
```
JeroenDeDauw, line 64: instancePropertyId parameter is unused (and also does not make sense since you can $this->config->getInstanceOfId())

src/Application/QueryStringParser.php
```diff
@@ -38,11 +43,33 @@ private function splitQueryString( string $queryString ): array {
 		);
 	}
 
+	private function isInstancePart( string $part ): bool {
+		return str_starts_with( $part, 'haswbstatement:' );
```
JeroenDeDauw, line 47: We need a new constructor argument: `private readonly PropertyId $instanceOfId`

Then in here we need to only match `'haswbstatement:' . $this->instanceOfId->getSerialization()`

src/Presentation/UiBuilder.php
```diff
@@ -26,19 +27,27 @@ public function __construct(
...
+
 		return $this->renderTemplate(
+			$this->buildInstancesViewModel(
+				instancePropertyId: $query->getInstancePropertyId(),
+				instanceItemId: $instanceItemId
+			),
 			$this->buildFacetsViewModel(
-				itemType: new ItemId( 'Q5976449' ), // TODO: get from search string
-				query: $this->parseQuery( $searchQuery )
+				instanceItemId: $instanceItemId,
```
JeroenDeDauw, line 39: I think the prior naming `itemType` is more appropriate as it matches the domain language

src/Application/QueryStringParser.php
```diff
@@ -38,11 +43,33 @@ private function splitQueryString( string $queryString ): array {
...
 	private function isFacetPart( string $part ): bool {
 		return str_starts_with( $part, 'haswbfacet:' )
 			|| str_starts_with( $part, '-haswbfacet:' );
 	}
 
+	/**
+	 * @param array<string, NumericPropertyId|ItemId> $instance
+	 * @return array<string, NumericPropertyId|ItemId>
+	 */
+	private function handleInstancePart( string $part, array &$instance ): array {
```
JeroenDeDauw, line 59: `$instance` should not be by-reference

Name isn't great. What about `itemTypes`?

JeroenDeDauw: Parameter and return types should change to `ItemId[]`

JeroenDeDauw: Btw, the existing return type confused me for a while. First I wondered "why is this an array" and then I thought "because an item can have multiple types". But no, it's an array because you return two different things.

I suppose it is fine for the MVP to only look at the first item type. This will work for the first use case of the extension.

tests/phpunit/Application/QueryStringParserTest.php
```diff
@@ -49,6 +50,17 @@ public function testParsesNonExistenceConstraint(): void {
...
+	public function testParsesInstance(): void {
+		$parser = new QueryStringParser();
+		$query = $parser->parse( 'haswbstatement:P42=Q68' );
+
+		$propertyId = new NumericPropertyId( 'P42' );
+		$itemId = new ItemId( 'Q68' );
+
+		$this->assertEquals( $propertyId, $query->getInstancePropertyId() );
+		$this->assertEquals( $itemId, $query->getInstanceItemId() );
+	}
```
JeroenDeDauw, line 62: After this test is adjusted to provide `PropertyId $instanceOfId` to `new QueryStringParser`, some additional tests will be in order:

* testItemTypeIsNullWhenNotPresent
* testIgnoresHaswbstatementForNonInstanceOfIdProperties

and this test itself could be renamed to `testFindsFirstItemType` and use a query string like 

    unrelated haswbstatement:P42=Q68 haswbstatement:P42=Q69

so that the test verifies the precise behavior. For testIgnoresHaswbstatementForNonInstanceOfIdProperties you could have

    haswbstatement:P1=wrongId haswbstatement:P42=Q68 haswbstatement:P2=alsoWrong

src/Application/Query.php
alistair3149: Did not know that, will update it tomorrow

src/Presentation/UiBuilder.php
alistair3149: In that case, should we use `instanceOfId` or should we use `itemType`?

JeroenDeDauw: Both, those are different things:

instanceOfId is coming from config and is a property ID

itemType is coming from the query string and is the value an item needs to have for the instanceOfId property id for it to show up in the search

src/Application/QueryStringParser.php
JeroenDeDauw: Good to use the PropertyId type, since this thing is being passed around between multiple classes

We probably should put this into a try-catch, because the constructor will throw on invalid property IDs. This should have an accompanying unit test

tests/phpunit/Application/QueryStringParserTest.php
```diff
@@ -144,7 +146,7 @@ public function testHandlesSingleQuotedString(): void {
 	public function testHandlesOrQuotedStrings(): void {
 		$this->markTestSkipped( 'TODO' );
 
-		$parser = new QueryStringParser();
+		$parser = $this->newQueryStringParser();
```
JeroenDeDauw, line 168: And this is why it's often worth creating that private method right away, even if initially there are no constructor arguments. Take note @JeroenDeDauw

Issue #118, opened 2025-01-15
Facet UI: Add input validation for range facet


alistair3149:
AC:
- Invalidate input when minimum is greater than maximum, or vise-versa
- Disable apply button when input is invalidated
- Show error state when input is invalidated
- Update `min` and `max` attributes accordingly to current input values

JeroenDeDauw:
> Update min and max attributes accordingly to current input values

How so? Unclear to me what would be sensible mix/max attributes on the inputs

alistair3149:
> How so? Unclear to me what would be sensible mix/max attributes on the inputs

For example, when user inputs the value 3 in the maximum `<input>` element, then the minimum `<input>` element should have 3 as the `max` value so hat we can do constraint validation on the client-side

JeroenDeDauw:
Will that prevent the user from putting a higher number in min? Or will it merely highlight the field as invalid if they do?

alistair3149:
> Will that prevent the user from putting a higher number in min? Or will it merely highlight the field as invalid if they do?

 User can still put an invalid value into the field, but the `<input>` element will have an `:invalid`  pseudo-class and it can be used to show error state.

JeroenDeDauw:
Sounds good to me then

alistair3149:
Completed with #128

PR #128, opened 2025-01-16
Add input validation for range facet

alistair3149:
#118 

Key changes:
- Invalidate input when minimum is greater than maximum, or vise-versa
- Disable apply button when input is invalidated
- Show error state when input is invalidated
- Update min and max attributes accordingly to current input values

### DIFF
--- a/resources/ext.wikibase.facetedsearch.js
+++ b/resources/ext.wikibase.facetedsearch.js
@@ -78,7 +78,7 @@ function onRangeFacetInput( facet, propertyId ) {
 	updateErrorState( minInput );
 	updateErrorState( maxInput );
 
-	if ( minInput.validity.valid === false || maxInput.validity.valid === false ) {
+	if ( !minInput.validity.valid || !maxInput.validity.valid ) {
 		applyButton.disabled = true;
 		return;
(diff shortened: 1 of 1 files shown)


resources/ext.wikibase.facetedsearch.js
```diff
@@ -73,6 +73,16 @@ function onRangeFacetInput( facet, propertyId ) {
 		return;
 	}
 
+	minInput.max = maxInput.value || '';
+	maxInput.min = minInput.value || '';
+	updateErrorState( minInput );
+	updateErrorState( maxInput );
+
+	if ( minInput.validity.valid === false || maxInput.validity.valid === false ) {
```
malberts, line 81: I think it's unnecessary to have explicit checks on `validity.valid`, since it's already defined as a boolean.
Screenshot_20250117_002128

JeroenDeDauw: I clarified that a few days ago in our Coding Conventions because I spotted a similarly redundant check in PHP

image

JeroenDeDauw (commented):
Looks good besides the commented on code

PR #131, opened 2025-01-17
Making naming consistent

JeroenDeDauw:
Partially a follow up to
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/127

We now have
* `InstanceOf id` for the property IDs
* `item type` for the values (item IDs)

`InstanceOf id` because the Wikibase community calls it the "InstanceOf property". It does read a bit weird, if you do not know that "InstanceOf" is one concept. I'm also wondering about the "ID" part: yes, it is an ID, but it is a pointer to a property. How about changing to something like "itemTypeProperty" in a follow up?


src/Application/Query.php
```diff
@@ -34,7 +34,10 @@ public function getFreeText(): string {
 		return $this->freeText;
 	}
 
-	public function getInstanceItemTypes(): array {
+	/**
+	 * @return ItemId[]
+	 */
+	public function getItemTypes(): array {
```
JeroenDeDauw, line 40: @alistair3149 we call this code, so the type should be there

tests/phpunit/Application/QueryStringParserTest.php
```diff
@@ -177,15 +198,9 @@ public function testHandlesOrQuotedStrings(): void {
...
-	private function newQueryStringParser(): QueryStringParser {
+	private function newQueryStringParser( ?string $instanceOfId = 'P11111' ): QueryStringParser {
 		return new QueryStringParser(
-			instanceType: $this->newConfig()->getInstanceOfId()
-		);
-	}
-
-	private function newConfig(): Config {
-		return new Config(
-			instanceOfId: new NumericPropertyId( self::INSTANCE_TYPE_ID )
```
JeroenDeDauw, line 188: Unsure why we built a Config just to get the ID

tests/phpunit/Application/QueryStringParserTest.php
```diff
@@ -53,33 +50,57 @@ public function testParsesNonExistenceConstraint(): void {
...
 
-		$this->assertEquals( [], $query->getInstanceItemTypes() );
+		$this->assertEquals( $itemTypes, $query->getItemTypes() );
 	}
 
 	public function testIgnoresHaswbstatementForNonInstanceOfIdProperties(): void {
-		$parser = $this->newQueryStringParser();
-		$query = $parser->parse( 'haswbstatement:P1=wrongId haswbstatement:' . self::INSTANCE_TYPE_ID . '=Q68 haswbstatement:P2=alsoWrong' );
+		$parser = $this->newQueryStringParser( instanceOfId: 'P32' );
+		$query = $parser->parse( 'haswbstatement:P1=wrongId haswbstatement:P32=Q68 haswbstatement:P2=alsoWrong' );
```
JeroenDeDauw, line 86: I find this more readable than using a constant. By proving the ID in the line above, it is obvious what the expected ID is, and the query remains easy to read.

tests/phpunit/Application/QueryStringParserTest.php
alistair3149: I thought we have to replicate the constructor like the original class for tests. Is simplification without the Config acceptable?

tests/phpunit/Application/QueryStringParserTest.php
alistair3149: That works too. On a side note, would it be more readable if the query string is broken down into an array and join together as a string?

src/WikibaseFacetedSearchExtension.php
```diff
@@ -147,9 +147,9 @@ private function newDataValueTranslator(): DataValueTranslator {
 		return new DataValueTranslator();
 	}
 
-	private function newInstanceTypeExtractor(): InstanceTypeExtractor {
-		return new InstanceTypeExtractor(
-			instanceType: $this->getConfig()->getInstanceOfId()
+	private function newInstanceTypeExtractor(): ItemTypeExtractor {
```
alistair3149, line 150: Should we rename `newInstanceTypeExtractor()` too?

src/Application/StatementListTranslator.php
```diff
@@ -13,7 +13,7 @@ class StatementListTranslator {
 
 	public function __construct(
 		private readonly StatementTranslator $statementTranslator,
-		private readonly InstanceTypeExtractor $instanceTypeExtractor,
+		private readonly ItemTypeExtractor $instanceTypeExtractor,
```
alistair3149, line 16: Should we rename `$instanceTypeExtractor` to `$itemTypeExtractor`?

tests/phpunit/Application/QueryStringParserTest.php
JeroenDeDauw: If the string are a bunch longer than perhaps a solution like that becomes worthwhile. Currently seems like a net negative

JeroenDeDauw:
Rebased and fixed the renames I missed

PR #133, opened 2025-01-19
Fix installMediaWiki.sh

JeroenDeDauw:
We did not see it fail on this repo yet due to the cache

You can see the issue at https://github.com/ProfessionalWiki/Maps/commits/master/


malberts:
IIRC this issue only happens when running CI with old versions of MW (<1.39).

Issue #136, opened 2025-01-19
Show item labels instead of Q-IDs


JeroenDeDauw:
In the list facets, show item labels instead of the Q-ID values

![Image](https://github.com/user-attachments/assets/2bb7c66c-2d1f-4f00-bc2b-adc64c2b4a28)

AC:
* Show Item label for the item referenced by the Q-ID value instead of that Q-ID
* Show in user language or fallback to wiki language (can be done in follow up)

JeroenDeDauw:
Implementation thoughts:

For each (list) facet, output the type of the property in a data attribute. You can get the type via Wikibase's `PropertyDataTypeLookup`. So the HTML would contain something like `data-property-type="wikibase-item"`.

Then use some JS to loop over the values in facets that have `wikibase-item` as Property type. Each value will be a Q-ID string. Once all Q-ID values are collected from the page, make a request to the API

    curl "https://www.wikidata.org/w/api.php?action=wbgetentities&ids=Q42|Q1337|Q9001&props=labels&languages=en&format=json"

```json
{
  "entities": {
    "Q42": {
      "type": "item",
      "id": "Q42",
      "labels": {}
    },
    "Q1337": {
      "type": "item",
      "id": "Q1337",
      "labels": {
        "en": {
          "language": "en",
          "value": "leet"
        }
      }
    },
    "Q9001": {
      "type": "item",
      "id": "Q9001",
      "labels": {
        "en": {
          "language": "en",
          "value": "Isle of Man Airport"
        }
      }
    }
  },
  "success": 1
}
```

Then loop over the response and update the shown values in the facets. We do need to keep track of the actual values to correctly update the query, so presumably we want to introduce some data-value attribute that always holds the Q-ID (at least for list facets for properties of type wikibase-item) and is used by the query building code.

This might already implement language fallbacks. Wikibase has some fancy support in its backend that ideally we use. I did not check if we need to use specific API request parameters to benefit from that, such as omitting the language.

Alternatively everything can be done in the backend, but this seems a nice case for progressive enhancement.

alistair3149:
The labels can be fetched through JS but I'm not sure if we have a need to do it JS as opposed to in the backend when we are reloading the page.

We already build the whole Facet within the backend. Having the label in JS will introduce a loading delay switching from the Q-values to the translated label. Depending on network conditions, it can give user the impressions that the UI is slow because of the lazyloading.

JeroenDeDauw:
[...]

Issue #107, opened 2025-01-14
[Low prio] Fix Config design issues


JeroenDeDauw:
We have some design issues with Config:
* It is used for "config building" and during runtime. These two cases are best served by different code. For config building, everything is optional. While during runtime values are required and we want an already item type filtered FacetConfigList indexed by property id.
* ConfigDeserializer return default config on validation failure. This makes it annoying to work with, since validtion failures are silently ignored.

PR #137, opened 2025-01-20
Use localized label instead of entity IDs

alistair3149:
#136

Key changes:
- Use labels when available for entity IDs on the facets
- Correct `templatePsy` spelling to `templateSpy`

Follow-up:
- Allow label from config to override the tab label (might be blocked by #107)


JeroenDeDauw:
This looks better and is closer to what we want on a product level, but it seems to not move us closer solution/code wise to the minimum acceptable behavior. I suspect this is slightly net negative by making getting to a minimum acceptable behavior slightly more work by having to remove the code for this intermediate state.

JeroenDeDauw:
I just realized you are also using this for the property IDs in the facets. So disregard my previous comment

PR #139, opened 2025-01-20
Reformat for readability

JeroenDeDauw:


tests/jest/ext.wikibase.facetedsearch.test.js
```diff
@@ -51,12 +51,22 @@ describe( 'ext.wikibase.facetedsearch.js', () => {
...
 	test( 'Query string with new instance query', () => {
-		expect( actual.buildQueryString( 'freetext haswbstatement:P1=Q10 haswbfacet:P2=Q20 haswbfacet:P3=Q30', [ 'haswbstatement:P1=Q11' ] ) )
-			.toEqual( 'freetext haswbstatement:P1=Q11' );
+		expect(
+			actual.buildQueryString(
+				'freetext haswbstatement:P1=Q10 haswbfacet:P2=Q20 haswbfacet:P3=Q30',
+				[ 'haswbstatement:P1=Q11' ],
+				'P404'
+			)
+		).toEqual( 'freetext haswbstatement:P1=Q11' );
```
JeroenDeDauw, line 60: TBH I am confused about this test. Maybe it's too late for me. Will check again tomorrow.

It's unclear to me why `haswbfacet:P2=Q20 haswbfacet:P3=Q30` is being dropped.

`propertyId` is also not a good name for the parameter in these methods. That tells us the type but not what it means.

tests/jest/ext.wikibase.facetedsearch.test.js
```diff
@@ -51,12 +51,21 @@ describe( 'ext.wikibase.facetedsearch.js', () => {
 	} );
 
 	test( 'Query string with new instance query', () => {
-		expect( actual.buildQueryString( 'freetext haswbstatement:P1=Q10 haswbfacet:P2=Q20 haswbfacet:P3=Q30', [ 'haswbstatement:P1=Q11' ] ) )
-			.toEqual( 'freetext haswbstatement:P1=Q11' );
+		expect(
+			actual.buildQueryString(
+				'freetext haswbstatement:P1=Q10 haswbfacet:P2=Q20 haswbfacet:P3=Q30',
+				[ 'haswbstatement:P1=Q11' ]
+			)
```
JeroenDeDauw, line 58: I notice the third argument is missing in this test

tests/jest/ext.wikibase.facetedsearch.test.js
JeroenDeDauw: Ah, `propertyId` is optional and `getFilteredQueries` has different behavior if it is set. It's being used by the tabs, which do not have this propertyId set, to remove a facet query segments when switching between tabs.

PR #140, opened 2025-01-21
Improve UiBuilder tests

JeroenDeDauw:


src/Presentation/UiBuilder.php
```diff
@@ -12,14 +12,14 @@
...
-use Wikibase\Lib\Store\FallbackLabelDescriptionLookup;
+use Wikibase\DataModel\Services\Lookup\LabelLookup;
 
 class UiBuilder {
 
 	public function __construct(
 		private readonly Config $config,
 		private readonly FacetHtmlBuilder $facetHtmlBuilder,
-		private readonly FallbackLabelDescriptionLookup $labelDescriptionLookup,
+		private readonly LabelLookup $labelLookup,
```
JeroenDeDauw, line 22: @malberts @alistair3149 It's important to use the most narrow interface

malberts: The Fallback version was required originally because the search rewrite had to include the Item Description (with language Fallback). After https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/113 it was no longer necessary.

JeroenDeDauw: Right ok

I suspect that in that case you'd still want LabelLookup and a DescriptionLookup

PR #138, opened 2025-01-20
Add unit tests for JS query string

alistair3149:
Closes #112

PR #141, opened 2025-01-21
Follow up to #138

alistair3149:
Key changes:
- Group test functions by their original function
- Add Jest to Makefile and document Jest command usage

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,8 +23,11 @@ stan:
 stan-baseline:
 	vendor/bin/phpstan analyse --configuration=phpstan.neon --memory-limit=2G --generate-baseline
 
+npm-install:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -25,3 +25,6 @@ stan-baseline:
 
 lint-docker:
 	docker run -it --rm -v "$(CURDIR)":/home/node/app -w /home/node/app -u node node:20 npm install && npm run lint
+
+jest:
+	npm install && npm run test
```
JeroenDeDauw, line 30: Should be via docker like the above command

Also should not include npm install. We can have a separate make command for that, but it should not be included when running `make jest` to avoid slowing down test feedback

JeroenDeDauw: By using Docker, we avoid needing local PHP, NPM, etc, and associated potential version or config issues

PR #142, opened 2025-01-22
Index item type

malberts:
Index the item type to allow it to be used for filtering results by type, using `haswbfacet`.

image

### DIFF
--- a/src/Application/StatementListTranslator.php
+++ b/src/Application/StatementListTranslator.php
@@ -30,6 +30,7 @@ class StatementListTranslator {
 		}
 
 		$propertyIds = $this->getPropertiesToIndex( $itemType );
+		$propertyIds[] = $this->config->getInstanceOfId();
 
 		$values = [];
 
--- a/src/Persistence/Search/SearchIndexFieldsBuilder.php
+++ b/src/Persistence/Search/SearchIndexFieldsBuilder.php
@@ -25,20 +25,25 @@ class SearchIndexFieldsBuilder {
 	 * @return array<string, SearchIndexField>
 	 */
 	public function createFields(): array {
-		$fields = [];
-
-		foreach ( $this->config->getFacets()->asArray() as $facetConfig ) {
-			$fieldType = $this->getFieldTypeForPropertyId( $facetConfig->propertyId );
-
-			if ( $fieldType === null ) {
-				continue;
-			}
-
-			$name = $this->getFacetFieldName( $facetConfig );
-			$fields[$name] = $this->makeSearchFieldMapping( $name, $fieldType );
+		try {
+			$itemType = $this->config->getInstanceOfId();
+		} catch ( Exception ) {
+			return [];
 		}
 
-		return $fields;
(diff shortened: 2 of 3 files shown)


src/Persistence/Search/SearchIndexFieldsBuilder.php
```diff
@@ -25,20 +25,25 @@ public function __construct(
...
 	public function createFields(): array {
-		$fields = [];
-
-		foreach ( $this->config->getFacets()->asArray() as $facetConfig ) {
-			$fieldType = $this->getFieldTypeForPropertyId( $facetConfig->propertyId );
+		try {
+			$itemType = $this->config->getInstanceOfId();
+		} catch ( Exception ) {
+			return [];
+		}
```
malberts, line 32: If InstanceOfId is missing, we throw an exception. However, this will cause indexing to break, for example if the user enabled this extension but did not start using it yet.
On the other hand, if InstanceOfId is missing, then there is probably no point in indexing anything else, because our tabs will not be able to filter by type.

src/Application/StatementListTranslator.php
```diff
@@ -30,6 +30,7 @@ public function translateStatements( StatementList $statements ): array {
 		}
 
 		$propertyIds = $this->getPropertiesToIndex( $itemType );
+		$propertyIds[] = $this->config->getInstanceOfId();
```
JeroenDeDauw, line 33: Putting it here makes `getPropertiesToIndex` lie

src/Persistence/Search/SearchIndexFieldsBuilder.php
JeroenDeDauw: Checking for the exception is weird.

`$this->config->hasInstanceOfId()` is better

But why bind to that detail. Could do `$this->config->isComplete()` or similar

Going further, maybe this service should not know about config potentially incomplete enough to prevent the extension from functioning. Maybe the code that executes this service could only be registered if the config is complete? This could also apply to other parts of the extension.

src/Persistence/Search/SearchIndexFieldsBuilder.php
```diff
@@ -25,20 +25,25 @@ public function __construct(
 	 * @return array<string, SearchIndexField>
 	 */
 	public function createFields(): array {
-		$fields = [];
-
-		foreach ( $this->config->getFacets()->asArray() as $facetConfig ) {
-			$fieldType = $this->getFieldTypeForPropertyId( $facetConfig->propertyId );
+		try {
```
JeroenDeDauw, line 28: `createFieldObjects` might be better. Generally words like `Data`, `Object`, and `Info`, are to be avoided, but in this case you make it clear that you are not creating the field in elastic.

JeroenDeDauw: The return type doc could say something about the key being the name for the index.

Could also create a value object and return `ThatNewType[]`, but not a huge deal

JeroenDeDauw (approved):

Issue #143, opened 2025-01-22
Show tabs on empty search


malberts:
The tabs are missing when:
* going to `Special:Search` directly
* doing an empty search (via `Special:Search` or the search form on all pages)

![Image](https://github.com/user-attachments/assets/3dfd66b4-ff39-4f12-bf42-8701cc4d0c57)

malberts:
We might have to use the `SpecialSearchResultsPrepend` hook for the tabs.

The `SpecialSearchResultsAppend` [docs](https://www.mediawiki.org/wiki/Manual:Hooks/SpecialSearchResultsAppend) mention:

> Called after all search results HTML has been output. Note that in some cases, this hook will not be called (no results, too many results, SpecialSearchResultsPrepend returned false, etc).

PR #145, opened 2025-01-22
Switch to SpecialSearchResultsPrepend hook so that Facet UI is always…

alistair3149:
… avaliable

Closes: #143


JeroenDeDauw:
Confirmed locally

PR #146, opened 2025-01-23
Show item label instead of Q-IDs if possible

alistair3149:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/136

Key changes
- Try to look up localized label for facet items


tests/phpunit/Application/LocalizedTextLookupTest.php
```diff
@@ -0,0 +1,59 @@
...
+
+	protected function setUp(): void {
+		$this->repo = $this->newMockRepository();
+	}
+
+	public function testGetLabelFromEntityIdString() {
+		$lookup = $this->newLocalizedTextLookup();
+
+		$this->setLabelToEntity( 'Test label', 'Q1' );
+		// TODO: Might have to mock labelLookup or figure out a way to use Wikibase
```
alistair3149, line 28: @JeroenDeDauw I am not sure what approach to take when it comes to testing getting labels from Wikibase entities. In my pursuit for answer in the Wikibase documentations, every pages have your name on it so I figure that you might know what would be a good approach 😅

image

src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -102,7 +104,7 @@ private function buildCheckboxesViewModel( FacetConfig $config, PropertyConstrai
 
 		foreach ( $this->getValuesAndCounts( $config ) as $i => $valueCount ) {
 			$checkboxes[] = [
-				'label' => $valueCount->value,
+				'label' => $this->localizedTextLookup->getLabelFromEntityIdString( (string)$valueCount->value ),
```
JeroenDeDauw, line 107: This is wrong. There is no guarantee that the value is an item id. The [logic that looks at the type of the property of the facet that I described for the frontend approach is still needed](https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/136#issuecomment-2601041463).

tests/phpunit/Application/LocalizedTextLookupTest.php
JeroenDeDauw: There is no need to create entities or persist anything

src/Presentation/ListFacetHtmlBuilder.php
JeroenDeDauw, line 107: My first reaction to injecting `LocalizedTextLookup` into the constructor of this class is that we should use `LabelLookup` instead. `ListFacetHtmlBuilder` does not need to know that localization is happening. That is an implementation detail of the collaborator.

Second thought: because `$valueCount->value` is a string, we either need to use a new type like the one you added, or use the `EntityIdParser` here so we can give an `EntityId` to `LabelLookup`. Putting the `EntityIdParser` into `ListFacetHtmlBuilder` is meh. Both approaches are meh.

Third thought: we do not need `EntityIdParser`. That class is mainly for cases where you do not know the ID type. If we go from an Item ID serialization to an Item ID object, we can simply do `new ItemId( $idString )`. Still need a try-catch, but this is less meh than needing `EntityIdParser` in `ListFacetHtmlBuilder`.

Conclusion: `ListFacetHtmlBuilder` should take `LabelLookup`. `LocalizedTextLookup` should implement `LabelLookup`.

#iAmNowAChainOfThoughtModel

JeroenDeDauw: Hmm... `LabelLookup` isn't the best interface... My above suggestion might not be better than the current state after all because `LocalizedTextLookup` provides an interface that makes more sense here. New suggestion: do not change current approach.

JeroenDeDauw: The above-mentioned logic issue seems to change the interface you need anyway.

`FacetLabelBuilder->buildLabel( string $valueCount->value, PropertyId $config->propertyId )`

And in `FacetLabelBuilder` (renamed from `LocalizedTextLookup`) you will need a `PropertyDataTypeLookup` collaborator to find out if you are dealing with item id values.

JeroenDeDauw: Though like that we get one call to the `PropertyDataTypeLookup` per value. Not necessarily an issue if the used `PropertyDataTypeLookup` implementation does some caching, but certainly not elegant. `FacetLabelBuilder` could do a simple cache itself with a `private readonly array $propertyIdToType` map.

alistair3149: When I create the `LocalizedTextLookup ` class, I was thinking about having a class to get localized text for the whole extension, since the search result title requires it too: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/blob/e14db07a8e64aa5ee27d9a7ec2c1546cfde32995/src/EntryPoints/WikibaseFacetedSearchHooks.php#L50

Or should we have `FacetLabelBuilder` for facet label and keep the search result title as the way it was?

tests/phpunit/Application/LocalizedTextLookupTest.php
alistair3149: It does, thanks for sharing the issue!

src/Presentation/ListFacetHtmlBuilder.php
JeroenDeDauw: I'm unsure what the linked code in the hooks file does (where in the UI it is used).

Having a `LocalizedTextLookup` seems plausible, especially if we have use for it in multiple places. We still need a `FacetLabelBuilder` here, which could have a `LocalizedTextLookup` collaborator, but would also contain other logic such as the calling of `PropertyDataTypeLookup`

alistair3149: > I'm unsure what the linked code in the hooks file does (where in the UI it is used).

It is used for the search result title in #113. Apart from that, we might need to use localized text outside of facets so it will be helpful to have a separate class for it. I will look into `FacetLabelBuilder` and see what we can do on the caching side.

JeroenDeDauw: > see what we can do on the caching side

The map approach that I described fully fixes the issue and is much simpler than using caching services, so please don't overcomplicate things.

alistair3149: @JeroenDeDauw would you be able to take a look into the current patch? I haven't updated the tests yet but I will follow up with that on another day.

I have created the `FacetLabelBuilder` class to handle all i18n labels needed by the facets. The class will then call `ItemTypeLabelLookup` (which replaced `LocalizedTextLookup` in my implementation, since they did the same thing) if the label can be retrieved from Wikibase (e.g. the value is a Wikibase entity). If the approach is fine, then I will update the test and rename the class to be more generic.

> The map approach that I described fully fixes the issue and is much simpler than using caching services, so please don't overcomplicate things.

I have implemented it as a part of the `FacetLabelBuilder` class.

src/Presentation/FacetLabelBuilder.php
```diff
@@ -0,0 +1,57 @@
...
+
+use ProfessionalWiki\WikibaseFacetedSearch\Application\ItemTypeLabelLookup;
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Services\Lookup\PropertyDataTypeLookup;
+use Wikibase\DataModel\Services\Lookup\PropertyDataTypeLookupException;
+
+class FacetLabelBuilder {
+
+	private array $propertyIdToType;
```
JeroenDeDauw, line 15: A type doc here is nice

src/Presentation/FacetLabelBuilder.php
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+		try {
+			$dataTypeId = $this->dataTypeLookup->getDataTypeIdForProperty( $propertyId );
+		} catch ( PropertyDataTypeLookupException ) {
+			return null;
+		}
+
+		$this->propertyIdToType[$key] = $dataTypeId;
+		return $this->propertyIdToType[$key];
```
JeroenDeDauw, line 55: Can be simplified to

```php

if ( !array_key_exists( $key, $this->propertyIdToType ) ) {
			$this->propertyIdToType[$key] = $this->getDataTypeIdForProperty( $propertyId );
		}

return $this->propertyIdToType[$key];
```

plus

```php
function getDataTypeIdForProperty() {
try {
			return $this->dataTypeLookup->getDataTypeIdForProperty( $propertyId );
		} catch ( PropertyDataTypeLookupException ) {
			return null;
		}
}
```

JeroenDeDauw: This also has the benefit of storing `null`, making the cache remember cases where there is no type.

src/Application/ItemTypeLabelLookup.php
```diff
@@ -4,10 +4,11 @@
...
 namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
 
-use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Entity\EntityId;
 
+// TODO: Make into a generic class that can be used for other labels
 interface ItemTypeLabelLookup {
 
-	public function getLabel( ItemId $itemType ): string;
+	public function getLabel( EntityId $itemType ): string;
```
JeroenDeDauw, line 12: This interface is designed for its existing consumer. Using it and its implementation of `FallbackItemTypeLabelLookup` elsewhere for other use cases (generic Wikibase Item label lookup and Property label lookup) isn't good.

src/Presentation/UiBuilder.php
```diff
@@ -105,16 +101,12 @@ private function buildFacetsViewModel( ?ItemId $itemType, Query $query ): array
 
 	private function buildFacetViewModel( FacetConfig $facet, PropertyConstraints $state ): array {
 		return [
-			'label' => $this->getPropertyLabel( $facet->propertyId ),
+			'label' => $this->facetLabelBuilder->getFacetLabel( $facet->propertyId ),
```
JeroenDeDauw, line 104: I think the prior code is better.

Services normally cater to one use case. It's weird to have multiple public methods, as `FacetLabelBuilder` now has. 

The previous property label code works, is simple, and is tested. Perhaps the little logic it contains can be de-duplicated if you create a generic improved version of `LabelLookup` for other parts of the code. That generic improved version "BetterLabelLookup" could do the fallback to the ID serialization and return `string ` instead of `Term`. Unless you are already creating such an interface anyway for other code, it does not seem worth it to touch the existing code here.

JeroenDeDauw: Suggestion: drop `getFacetLabel` from `facetLabelBuilder`

src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -102,7 +103,7 @@ private function buildCheckboxesViewModel( FacetConfig $config, PropertyConstrai
 
 		foreach ( $this->getValuesAndCounts( $config ) as $i => $valueCount ) {
 			$checkboxes[] = [
-				'label' => $valueCount->value,
+				'label' => $this->labelBuilder->getItemLabel( (string)$valueCount->value, $config->propertyId ),
```
JeroenDeDauw, line 106: Our use of `label` here is unfortunate, as it creates confusion with Wikibase labels, which are not the same.

`Item` is similarly unfortunate, as it is not about Wikibase items.

This would be clearer: 
* `private FacetValueFormatter $valueFormatter`
* `'formattedValue' => $this->valueFormatter->formatValue( (string)$valueCount->value, $config->propertyId ),`

I'm not in love with that naming but it removes the ambiguity

src/Presentation/UiBuilder.php
```diff
@@ -63,7 +59,7 @@ private function buildTabsViewModel( ?ItemId $selectedItemType ): array {
 
 		foreach ( $this->config->getItemTypes() as $itemType ) {
 			$tabs[] = [
-				'label' => $this->itemTypeLabelLookup->getLabel( $itemType ),
+				'label' => $this->facetLabelBuilder->getTabLabel( $itemType ),
```
JeroenDeDauw, line 62: Same as the below comment. Old code was better, and `FacetLabelBuilder` should not have `getTabLabel`

src/Presentation/ListFacetHtmlBuilder.php
alistair3149: The use of `value` might be confused with the `value` attribute in `<input>` elements in HTML. What do you think about `labelText`?

src/WikibaseFacetedSearchExtension.php
```diff
@@ -253,4 +264,12 @@ public function getLabelLookup( Language $language ): LabelLookup {
 		return WikibaseRepo::getFallbackLabelDescriptionLookupFactory()->newLabelDescriptionLookup( $language );
 	}
 
+	public function getEntityIdParser(): EntityIdParser {
+		return WikibaseRepo::getEntityIdParser();
+	}
```
JeroenDeDauw, line 269: This seems unused

tests/phpunit/Presentation/FacetLabelBuilderTest.php
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * @covers \ProfessionalWiki\WikibaseFacetedSearch\Presentation\FacetLabelBuilder
+ */
+class FacetLabelBuilderTest extends TestCase {
+
+	public function testGetLabel(): void {
+		$facetLabelBuilder = $this->newFacetLabelBuilder();
+
+		$this->assertSame( 'Q100', $facetLabelBuilder->getLabel( 'Q100', new NumericPropertyId( 'P123' ) ) );
+	}
```
JeroenDeDauw, line 23: There are many test cases missing

src/Presentation/UiBuilder.php
```diff
@@ -105,16 +104,12 @@ private function buildFacetsViewModel( ?ItemId $itemType, Query $query ): array
 
 	private function buildFacetViewModel( FacetConfig $facet, PropertyConstraints $state ): array {
 		return [
-			'label' => $this->getPropertyLabel( $facet->propertyId ),
+			'label' => $this->labelLookup->getLabel( $facet->propertyId )?->getText() ?? $facet->propertyId->getSerialization(),
```
JeroenDeDauw, line 107: Bit weird this got inlined

src/Presentation/ListFacetHtmlBuilder.php
JeroenDeDauw: Ah, you are thinking about the naming from the HTML perspective. Unfortunate that the text there is called label, overloading that one term with multiple meanings.

Not immediately obvious to me if we should go with the domain langue (Wikibase) (my previous suggestion) or with the technical language (HTML) (your take). Ideally we avoid confusion regardless of the perspective.

Case of #NamingIsHard. Too hard for me during late evening

malberts: My first thought is that I lean slightly towards HTML language here, since this is ultimately about building a model for the frontend. And it's kind of "scoped" to a checkbox. So when I'm thinking about "the label of a checkbox" I think about what it means in the context of a checkbox, not in terms of what was used to create the "checkbox label". Although `getItemLabel` is indeed ambiguous, because it is actually used to get the WB _label_ of a WB _item_, in order to set the checkbox _label_ of a list _item_.

However, the "formatter" idea might actually be better for a different reason: there is a possibility that we will need to format numbers and dates when they are used in a list filter.

Right now dates show up as unreadable timestamps:
Screenshot_20250130_182334

And larger numbers might also require locale-specific formatting (e.g. thousands separator):
Screenshot_20250130_182420

alistair3149: After seeing @malberts' comment and I agree that the value formatter is a more suitable approach. We do need to format other types of values into human readable formats. I will refactor the PR accordingly.

PR #147, opened 2025-01-23
Rename instanceOf related symbols

JeroenDeDauw:
You will have to update your LocalSettings.php (or `MediaWiki:WikibaseFacetedSearch` if you set stuff there)

change

```json
	"instanceOfId": "P1460",
	"instanceOfValues": {
```

to

```json
	"itemTypeProperty": "P1460",
	"configPerItemType": {
```

### DIFF
--- a/src/Application/FacetConfig.php
+++ b/src/Application/FacetConfig.php
@@ -13,7 +13,7 @@ class FacetConfig {
 	 * @param array<string, mixed> $typeSpecificConfig
 	 */
 	public function __construct(
-		public readonly ItemId $itemTypeId,
+		public readonly ItemId $itemType,
 		public readonly PropertyId $propertyId,
 		public readonly FacetType $type,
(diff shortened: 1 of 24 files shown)


src/Application/FacetConfig.php
```diff
@@ -13,7 +13,7 @@ class FacetConfig {
 	 * @param array<string, mixed> $typeSpecificConfig
 	 */
 	public function __construct(
-		public readonly ItemId $itemTypeId,
+		public readonly ItemId $itemType,
 		public readonly PropertyId $propertyId,
```
malberts, line 17: I assume we'll do a similar refactor to change `$propertyId` to `$property` here?

JeroenDeDauw: Unclear that would be better

`itemTypeId vs itemType` and `propertyId vs property` are not equivalent

PR #149, opened 2025-01-23
Index item type refactor

malberts:
Follow-up to #142

### DIFF
--- a/tests/phpunit/Persistence/Search/SearchIndexFieldsBuilderTest.php
+++ b/tests/phpunit/Persistence/Search/SearchIndexFieldsBuilderTest.php
@@ -30,10 +30,12 @@ class SearchIndexFieldsBuilderTest extends TestCase {
 		$this->cirrusSearch = new CirrusSearch();
 	}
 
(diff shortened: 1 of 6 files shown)


tests/phpunit/Persistence/Search/SearchIndexFieldsBuilderTest.php
```diff
@@ -30,10 +30,12 @@ public function setUp(): void {
...
 	}
 
-	public function testEmptyConfigReturnsNoFields(): void {
+	public function testEmptyConfigThrowsException(): void {
 		$builder = $this->newBuilder( new Config() );
 
-		$this->assertSame( [], $builder->createFields() );
+		$this->expectException( \RuntimeException::class );
+
+		$builder->createFieldObjects();
```
malberts, line 38: I imagine to avoid this type of test where practically it should not happen (i.e. we're not constructing the object if Config is incomplete), but technically it can, we might have to inject the known values into SearchIndexFieldsBuilder instead of the whole Config object.

Issue #53, opened 2025-01-01
Generate (internal) elastic queries for haswbfacet


JeroenDeDauw:
Follow up to
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/38

Needs to parse the query string and might be able to use 
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/31

PR #152, opened 2025-01-24
Use haswbfacet for parsing item type

### LINKED ISSUE #150: Use haswbfacet instead of haswbstatement for query parser
Facets are indexed on Pages, and `haswbstatement` only works on Items.

Item type parsing should use `haswbfacet`.

AC:
* [ ] Update QueryStringParser

malberts:
Fixes #150 

Required by https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/53 to correctly detect the item type.

### DIFF
--- a/src/Application/QueryStringParser.php
+++ b/src/Application/QueryStringParser.php
@@ -51,7 +51,7 @@ class QueryStringParser {
 	}
 
 	private function isInstanceOfPart( string $part ): bool {
-		return str_starts_with( $part, 'haswbfacet:' . $this->itemTypeProperty->getSerialization() );
+		return str_starts_with( $part, 'haswbfacet:' . $this->itemTypeProperty->getSerialization() . '=' );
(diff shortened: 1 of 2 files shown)


src/Application/QueryStringParser.php
```diff
@@ -51,7 +51,7 @@ private function splitQueryString( string $queryString ): array {
 	}
 
 	private function isInstanceOfPart( string $part ): bool {
-		return str_starts_with( $part, 'haswbstatement:' . $this->itemTypeProperty->getSerialization() );
+		return str_starts_with( $part, 'haswbfacet:' . $this->itemTypeProperty->getSerialization() . '=' );
```
malberts, line 54: There was a logic issue here. Valid item type "P32" was going to be a substring of non-item type "P320".

JeroenDeDauw (approved):

Issue #58, opened 2025-01-02
Support more Property types


malberts:
Follow-up to:
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/49
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/56

AC:
* [ ] Determine which Property types should be supported
* [ ] Add more Property types for field creation: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/blob/master/src/Persistence/Search/SearchIndexFieldsBuilder.php
* [ ] Add more DataValue translations for data indexing: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/51/files#diff-3111d61a2aef9a5bce15e5497fd8f5e7ef81a793b73186b9f5d4923229b12b62

PR #153, opened 2025-01-24
Build Elasticsearch queries

malberts:
For #53 

### List facet
Supported data types:
* string
* item
* quantity
* time

Supported query types:
* AND
* OR
* has any value
* has no value

https://github.com/user-attachments/assets/929934d3-3f3c-4e8f-b2ff-aa6334e5319b

https://github.com/user-attachments/assets/64f94a5c-3ca3-4087-a802-28b9b5ad9f19

### Range facet:
Supported types:
* quantity
* time

https://github.com/user-attachments/assets/80e3eea8-7b23-4fac-956e-2e77f6e97292

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -20,6 +20,7 @@ parameters:
 		- src/Persistence/Search/SearchIndexFieldsBuilder.php
 		- src/Persistence/Search/Query/HasWbFacetFeature.php
 		- src/Persistence/Elastic*.php
+		- src/WikibaseFacetedSearchExtension.php
 	reportUnmatchedIgnoredErrors: false
 	ignoreErrors:
 		- '#no value type specified in iterable type array\.$#'
--- a/src/Persistence/Search/Query/HasWbFacetFeature.php
+++ b/src/Persistence/Search/Query/HasWbFacetFeature.php
@@ -4,14 +4,22 @@ declare( strict_types = 1 );
 
 namespace ProfessionalWiki\WikibaseFacetedSearch\Persistence\Search\Query;
 
-use CirrusSearch\Parser\AST\KeywordFeatureNode;
-use CirrusSearch\Query\Builder\QueryBuildingContext;
-use CirrusSearch\Query\FilterQueryFeature;
 use CirrusSearch\Query\SimpleKeywordFeature;
 use CirrusSearch\Search\SearchContext;
-use Elastica\Query\AbstractQuery;
+use ProfessionalWiki\WikibaseFacetedSearch\Application\Config;
+use ProfessionalWiki\WikibaseFacetedSearch\Application\PropertyConstraints;
+use ProfessionalWiki\WikibaseFacetedSearch\Application\QueryStringParser;
+use Wikibase\DataModel\Entity\ItemId;
 
-class HasWbFacetFeature extends SimpleKeywordFeature implements FilterQueryFeature {
+class HasWbFacetFeature extends SimpleKeywordFeature {
+
+	public function __construct(
+		private readonly Config $config,
+		private readonly QueryStringParser $queryStringParser,
+		private readonly ItemTypeQueryBuilder $itemTypeQueryBuilder,
+		private readonly FacetQueryBuilder $facetQueryBuilder
+	) {
+	}
 
 	/**
 	 * @return string[]
@@ -20,17 +28,46 @@ class HasWbFacetFeature extends SimpleKeywordFeature implements FilterQueryFeatu
 		return [ 'haswbfacet' ];
 	}
 
-	/**
-	 * @return array
-	 */
 	protected function doApply( SearchContext $context, $key, $value, $quotedValue, $negated ): array {
-		// TODO
+		$itemTypes = $this->getItemTypes( $context->getOriginalSearchTerm() );
+
+		if ( $itemTypes === [] ) {
+			return [ null, false ];
+		}
+
+		$constraints = $this->getConstraintsForValue( $value );
+
+		if ( $constraints === null ) {
+			return [ $this->itemTypeQueryBuilder->buildQuery( $itemTypes ), false ];
+		}
+
+		foreach ( $itemTypes as $itemTypeId ) {
+			$facet = $this->config->getConfigForProperty( $itemTypeId, $constraints->propertyId );
+
+			if ( $facet === null ) {
+				continue;
+			}
+
+			return [ $this->facetQueryBuilder->buildQuery( $facet, $constraints ), false ];
+		}
+
 		return [ null, false ];
 	}
 
-	public function getFilterQuery( KeywordFeatureNode $node, QueryBuildingContext $context ): ?AbstractQuery {
-		// TODO
-		return null;
+	/**
+	 * @return ItemId[]
+	 */
+	private function getItemTypes( string $originalQueryString ): array {
+		return $this->queryStringParser->parse( $originalQueryString )->getItemTypes();
+	}
+
+	/**
+	 * @param string $value Keyword value for a single facet
+	 */
+	private function getConstraintsForValue( string $value ): ?PropertyConstraints {
+		$constraints = $this->queryStringParser->parse( "haswbfacet:$value" )->getConstraintsPerProperty();
+
+		return $constraints === [] ? null : reset( $constraints );
 	}
 
 }
--- a/src/Persistence/Search/Query/ItemTypeQueryBuilder.php
+++ b/src/Persistence/Search/Query/ItemTypeQueryBuilder.php
@@ -1,0 +1,29 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Persistence\Search\Query;
+
+use Elastica\Query;
+use Elastica\Query\AbstractQuery;
+use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Entity\PropertyId;
+
+class ItemTypeQueryBuilder {
+
+	public function __construct(
+		private readonly PropertyId $itemTypeProperty
+	) {
+	}
+
+	public function buildQuery( array $itemTypes ): AbstractQuery {
+		return new Query\Terms(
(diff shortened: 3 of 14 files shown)


phpstan.neon
```diff
@@ -20,6 +20,7 @@ parameters:
 		- src/Persistence/Search/SearchIndexFieldsBuilder.php
 		- src/Persistence/Search/Query/HasWbFacetFeature.php
 		- src/Persistence/Elastic*.php
+		- src/WikibaseFacetedSearchExtension.php
```
malberts, line 23: This file got contaminated by CirrusSearch classes: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/44#discussion_r1888899011

src/Persistence/Search/Query/HasWbFacetFeature.php
```diff
@@ -20,17 +28,46 @@ protected function getKeywords(): array {
...
+
+		if ( $itemTypes === [] ) {
+			return [ null, false ];
+		}
+
+		$constraints = $this->getConstraintsForValue( $value );
+
+		if ( $constraints === null ) {
+			return [ $this->itemTypeQueryBuilder->buildQuery( $itemTypes ), false ];
+		}
```
malberts, line 42: This isn't very nice. We need to parse the full query string to get the current item type(s). But then we need to reparse just the current token to get the OR values, or a single AND value, since the actual query AND handling is done by the search engine (where each token's query gets AND-ed).

src/Persistence/Search/Query/HasWbFacetFeature.php
```diff
@@ -4,14 +4,22 @@
...
+use Wikibase\DataModel\Entity\ItemId;
 
-class HasWbFacetFeature extends SimpleKeywordFeature implements FilterQueryFeature {
+class HasWbFacetFeature extends SimpleKeywordFeature {
+
+	public function __construct(
+		private readonly Config $config,
+		private readonly QueryStringParser $queryStringParser,
+		private readonly ItemTypeQueryBuilder $itemTypeQueryBuilder,
+		private readonly DelegatingFacetQueryBuilder $facetQueryBuilder
```
JeroenDeDauw, line 20: Could be FacetQueryBuilder instead of DelegatingFacetQueryBuilder

src/Persistence/Search/Query/ItemTypeQueryBuilder.php
```diff
@@ -0,0 +1,29 @@
...
+
+class ItemTypeQueryBuilder {
+
+	public function __construct(
+		private readonly PropertyId $itemTypeProperty
+	) {
+	}
+
+	public function buildQuery( array $itemTypes ): AbstractQuery {
+		return new Query\Terms(
```
JeroenDeDauw, line 20: Is this partial import on purpose?

tests/phpunit/Persistence/Search/Query/RangeFacetQueryBuilderTest.php
```diff
@@ -0,0 +1,132 @@
...
+			self::TIME_PROPERTY => 'time',
+			self::ITEM_PROPERTY => 'wikibase-item'
+		];
+
+		foreach ( $types as $pId => $type ) {
+			$dataTypeLookup->setDataTypeForProperty(
+				new NumericPropertyId( $pId ),
+				$type
+			);
+		}
```
JeroenDeDauw, line 66: I suspect this would be clearer without the loop, but that is a minute CS difference

tests/phpunit/Persistence/Search/Query/RangeFacetQueryBuilderTest.php
```diff
@@ -0,0 +1,132 @@
...
+		$query = $this->newFacetQueryBuilder()->buildQuery(
+			$this->newRangeFacetConfig( self::ITEM_PROPERTY ),
+			new PropertyConstraints(
+				new NumericPropertyId( self::ITEM_PROPERTY ),
+				orSelectedValues: [ 'Q100', 'Q200' ]
+			)
+		);
+
+		$this->assertNull( $query );
+	}
```
JeroenDeDauw, line 130: There does not seem to be any test for the common-ish use case of open ranges (only min or only max)

src/Persistence/Search/Query/HasWbFacetFeature.php
JeroenDeDauw: Is this "just" a case of being wasteful by parsing too much?

Structurally the PR seems good. Some wasteful parsing probably has little or no real-world impact

src/Persistence/Search/Query/HasWbFacetFeature.php
```diff
@@ -20,17 +28,46 @@ protected function getKeywords(): array {
...
+		foreach ( $itemTypes as $itemTypeId ) {
+			$facet = $this->config->getConfigForProperty( $itemTypeId, $constraints->propertyId );
+
+			if ( $facet === null ) {
+				continue;
+			}
+
+			return [ $this->facetQueryBuilder->buildQuery( $facet, $constraints ), false ];
+		}
+
```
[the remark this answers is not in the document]
JeroenDeDauw: What are the benefits? Sounds plausible, but I don't immediately see what is gained (and idk what `AbstractQuery` is)

Currently this class seems sane enough, so I expect the gain to at most be going from good enough to good.

JeroenDeDauw (approved):

src/Persistence/Search/Query/ItemTypeQueryBuilder.php
alistair3149: Should the return type be `Terms` from `Elastica\Query\Terms` instead of `AbstractQuery`?

src/Persistence/Search/Query/RangeFacetQueryBuilder.php
```diff
@@ -0,0 +1,50 @@
...
+
+		return match ( $this->dataTypeLookup->getDataTypeIdForProperty( $config->propertyId ) ) {
+			'quantity' => $this->buildQuantityQuery( $name, $constraints ),
+			'time' => $this->buildTimeQuery( $name, $constraints ),
+			default => null
+		};
+	}
+
+	private function buildQuantityQuery( string $name, PropertyConstraints $constraints ): AbstractQuery {
+		return new Query\Range(
```
alistair3149, line 31: Same with above, should the return type be `Range` from  `Elastica\Query\Range` instead?

src/WikibaseFacetedSearchExtension.php
```diff
@@ -41,6 +46,7 @@
 use TemplateParser;
 use Title;
```
alistair3149, line 47: Not related to this change but we should update this to `MediaWiki\Title\Title`

tests/phpunit/Persistence/Search/Query/ListFacetQueryBuilderTest.php
```diff
@@ -0,0 +1,170 @@
...
+
+	private function newListFacetConfig( string $propertyId ): FacetConfig {
+		return new FacetConfig(
+			new ItemId( 'Q404' ),
+			new NumericPropertyId( $propertyId ),
+			FacetType::LIST
+		);
+	}
+
+	private function assertIsTermsQueryWithParams( string $propertyId, array $values, Query\AbstractQuery $query ): void {
```
alistair3149, line 77: Is the partial import intended?

tests/phpunit/Persistence/Search/Query/RangeFacetQueryBuilderTest.php
```diff
@@ -0,0 +1,132 @@
...
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Tests\Persistence\Search\Query;
+
+use Elastica\Query;
+use PHPUnit\Framework\TestCase;
+use ProfessionalWiki\WikibaseFacetedSearch\Application\FacetConfig;
+use ProfessionalWiki\WikibaseFacetedSearch\Application\FacetType;
+use ProfessionalWiki\WikibaseFacetedSearch\Application\PropertyConstraints;
+use ProfessionalWiki\WikibaseFacetedSearch\Persistence\Search\Query\DelegatingFacetQueryBuilder;
```
alistair3149, line 12: This class is unused

src/Persistence/Search/Query/HasWbFacetFeature.php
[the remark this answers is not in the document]
malberts: `AbstractQuery` is the abstract class for all Elastica query objects. And the expected first parameter of the returned array.

This would mainly be to test that query objects are (not) created based on what's in the search string. For example:
* if item type is missing, then no query gets creates
* if the current token is just the item type, then it returns an item type query
* otherwise returns the facet query

Maybe not a big gain.

src/Persistence/Search/Query/ItemTypeQueryBuilder.php
malberts: > Is this partial import on purpose?

It was kind of, since I this pattern being used in a few places for Elastica query objects. But I suppose this is arbitrary and WikibaseCirrusSearch imports them directly, so I'll do that instead.

malberts: > Should the return type be Terms from Elastica\Query\Terms instead of AbstractQuery?

No. All the query objects extend AbstractQuery. The calling code does not need to know what kind of query is constructed. It just needs to know the most narrow interface.

tests/phpunit/Persistence/Search/Query/RangeFacetQueryBuilderTest.php
malberts: I copied it from WikibaseExport: https://github.com/ProfessionalWiki/WikibaseExport/blob/d3b9f998fff2144e9f1ffd1a48040791917267aa/tests/Application/Export/ProductionValueSetCreatorTest.php#L44-L67

At some point we might need to define more types here: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/58

For future reference: is the concern here about having a loop for few items, or having a loop here in general?

Issue #156, opened 2025-01-25
Hide Filters on All tab


malberts:
There can never be filters on the All tab, so hide the Filters heading:
![Image](https://github.com/user-attachments/assets/f315e175-8d93-41e1-9e8f-32d987af61d2)

PR #159, opened 2025-01-27
Hide facets when there are no facets

alistair3149:
Closes: #156

### DIFF
--- a/src/Presentation/UiBuilder.php
+++ b/src/Presentation/UiBuilder.php
@@ -46,6 +46,7 @@ class UiBuilder {
 		return $this->templateParser->processTemplate(
 			'Layout',
 			[
+				'hasFacets' => count( $facetsViewModel ) > 0,
 				'instanceId' => $this->config->getItemTypeProperty()->getSerialization(),
 				'instances' => $instancesViewModel,
 				'facets' => $facetsViewModel,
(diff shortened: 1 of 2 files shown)


malberts:
Is it necessary to handle this in the template? Can we rather just not render the template at all?

alistair3149:
> Is it necessary to handle this in the template? Can we rather just not render the template at all?

Currently the Tabs template is included as a part of the layout template.
We can extract the Tabs template out so that Tabs and Facets are rendered separately, but that would be two calls to the `templateParser` and requires more logic in `UiBuilder`.

src/Presentation/UiBuilder.php
```diff
@@ -46,6 +46,7 @@ private function renderTemplate( array $instancesViewModel, array $facetsViewMod
 		return $this->templateParser->processTemplate(
 			'Layout',
 			[
+				'hasFacets' => count( $facetsViewModel ) > 0,
```
malberts, line 49: I think the intention will be more explicit as `showSidebar`. And then use that in `Layout` to skip rendering the `Sidebar` template. Instead of rendering the template, but with nothing inside. (Unless this causes a serious problem with the CSS layout right now.)

In the "All" view we will never have a sidebar. So theoretically if there was something else in the sidebar besides facets, then `hasFacets` wouldn't clearly be about (not) showing the sidebar.

PR #161, opened 2025-01-28
Add .vscode to gitignore

alistair3149:


malberts:
This should not be committed into the repo. Ideally IDE-specific ignores should go into your local `.gitignore_global`.

Issue #158, opened 2025-01-25
Make list facet "OR behavior" work


JeroenDeDauw:
AC:
* Can switch between AND and OR when both are enabled
* Builds correct query when OR is selected
* Updates query when switching from AND to OR or the other way round

PR #162, opened 2025-01-28
Add support for OR query in the Facet UI

alistair3149:
#158

### DIFF
--- a/resources/ext.wikibase.facetedsearch.js
+++ b/resources/ext.wikibase.facetedsearch.js
@@ -14,6 +14,7 @@ function init() {
 
 	if ( facets ) {
 		facets.addEventListener( 'input', onFacetsInput );
+		facets.addEventListener( 'click', onFacetsInput );
 
 		const
 			button = document.querySelector( '.wikibase-faceted-search__dialog-button' ),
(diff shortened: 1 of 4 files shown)


resources/ext.wikibase.facetedsearch.js
```diff
@@ -75,19 +79,32 @@ function onInstancesClick( event, instanceId ) {
...
 
+/**
+ * Determines the query mode for a list facet based on the selected toggle button.
+ *
+ * @param {HTMLDivElement} facet
+ * @return {string}
+ */
+function getListFacetQueryMode( facet ) {
+	const selectedButton = facet.querySelector( '.wikibase-faceted-search__facet-toggle > .cdx-button--action-progressive' );
+	return selectedButton ? selectedButton.value : 'AND';
```
JeroenDeDauw, line 100: Normally there should always be a selected button, right? In other words, this check is to deal with unexpected error states.

If not, and we can regularly have the state, then defauling to AND is likely incorrect behavior.

resources/ext.wikibase.facetedsearch.js
```diff
@@ -140,19 +157,33 @@ function updateErrorState( input ) {
...
+
+	const segments = [];
+	if ( mode === 'AND' ) {
+		checkedValues.forEach( ( value ) => {
+			segments.push( `haswbfacet:${ propertyId }=${ value }` );
+		} );
+	} else {
+		segments.push( `haswbfacet:${ propertyId }=${ checkedValues.join( '|' ) }` );
+	}
+
```
JeroenDeDauw, line 186: This function seems to be doing two different things: finding selected values in the DOM, and building query segments.

resources/ext.wikibase.facetedsearch.js
alistair3149: Yes normally there are always a selected button, this is used to deal with unexpected state as you mentioned

resources/ext.wikibase.facetedsearch.js
alistair3149: Yes that's right. I can refactor it into two functions in a follow up patch

JeroenDeDauw: Sounds good

Issue #79, opened 2025-01-05
Facets UI: Handle many values


JeroenDeDauw:
See the Author facet at https://zbmath.org/?q=a+sw%3AMatlab. It shows only the top 5, but allows showing more. The longer list still has a cutoff.

We could do something similar or better

AC:
* UI looks good even if there are 100 values provided by the backend (with 100 being the example cutoff the backend enforces in `ElasticValueCounter`)
* If there are 100 values, only show some at first, but allow viewing all

Ideally we also allow hiding the additional ones again.

This TBD task could build on the above:
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/78

![Image](https://github.com/user-attachments/assets/33399f7a-5eb0-479f-bab2-e63038f8ec4e)

## Inspiration

This is how Amazon does it:

![Image](https://github.com/user-attachments/assets/f3055a98-1335-4c45-b017-bf26442df17b)

![Image](https://github.com/user-attachments/assets/2432e778-26d6-42f9-b94b-a3b56bdd963b)

There is also https://zbmath.org/?q=wiki, which show the element at the top right.

alistair3149:
Moving the discussion from #157.

Are we lazyloading the facet item from the backend (e.g. API) or do it render all the facet items and hide the items in the frontend?

JeroenDeDauw:
No lazy loading. All 100 values are in the initial HTML. Presumably the initial HTML should already hide all but the top 5 values.

PR #164, opened 2025-01-29
Add overflow handling for list facets

alistair3149:
#79

https://github.com/user-attachments/assets/49af2585-1dbd-4bd2-9350-8987000d559b


src/Presentation/ListFacetHtmlBuilder.php
malberts, line 1: This works, but I am wondering about the approach. We are slightly complicating the backend with a client-side concern. I imagine the backend could just provide all the checkboxes, and then JS/CSS can be used to figure out what to show. Although then we'd need more complicated JS/CSS.

alistair3149: We can move the overflow rendering to the frontend. To do that, we need to hide the facets until the JS has set up the initial state of it. I prefer the PHP approach more because it is less render blocking, but the JS approach is fine as well.

tests/phpunit/Presentation/ListFacetHtmlBuilderTest.php
```diff
@@ -48,15 +56,33 @@ private function newListFacetHtmlBuilder(): ListFacetHtmlBuilder {
...
 
+	public function testCheckboxesViewModelContainsLabels(): void {
+		$viewModel = $this->buildViewModel();
+
+		$this->assertArrayHasKey( 'checkboxes', $viewModel );
+		$this->assertArrayHasKey( 'msg-show-more', $viewModel['checkboxes'] );
+		$this->assertArrayHasKey( 'msg-show-less', $viewModel['checkboxes'] );
+		$this->assertIsString( $viewModel['checkboxes']['msg-show-more'] );
+		$this->assertIsString( $viewModel['checkboxes']['msg-show-less'] );
+	}
```
JeroenDeDauw, line 67: This type of test adds little value because it is not testing logic. It is testing wiring, and doing so directly, on the unit test level. If the wiring is wrong, like using the wrong key, the test won't notice. An integration test that verifies the message is in the rendered template would provide some value, though it's not clear such a test is called for, certainly not if we are talking about multiple such tests for a single template rather than a smoke test.

src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -98,19 +98,33 @@ private function buildCheckboxesViewModel( FacetConfig $config, PropertyConstrai
...
 		}
 
-		return $checkboxes;
+		return [
+			'visible' => $visibleCheckboxes,
+			'collapsed' => $collapsedCheckboxes,
+			'showMore' => count( $collapsedCheckboxes ) > 0,
+			'msg-show-more' => wfMessage( 'wikibase-faceted-search-facet-show-more' )->text(),
+			'msg-show-less' => wfMessage( 'wikibase-faceted-search-facet-show-less' )->text()
+		];
```
JeroenDeDauw, line 127: Do we need to have those messages in here? Can't all messages in on the top level in `buildViewModel`?

src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -98,19 +98,33 @@ private function buildCheckboxesViewModel( FacetConfig $config, PropertyConstrai
...
 		foreach ( $this->getValuesAndCounts( $config ) as $i => $valueCount ) {
-			$checkboxes[] = [
+			$checkbox = [
 				'label' => $valueCount->value,
 				'count' => $valueCount->count,
 				'checked' => in_array( $valueCount->value, $selectedValues ), // TODO: test with multiple types of values
 				'value' => $valueCount->value,
 				'id' => $state->propertyId->getSerialization() . "-$i",
 			];
+
```
JeroenDeDauw, line 112: With the additional logic in this function, it becomes appealing to extract that body of the loop into a ne function

tests/phpunit/Presentation/ListFacetHtmlBuilderTest.php
```diff
@@ -106,9 +142,14 @@ public function testOnlyCheckboxesForValuesInThePropertyConstraintsAreChecked_or
...
-		$this->assertTrue( $viewModel['checkboxes'][1]['checked'] );
-		$this->assertFalse( $viewModel['checkboxes'][2]['checked'] );
+		$this->assertFalse( $viewModel['checkboxes']['visible'][0]['checked'] );
+		$this->assertTrue( $viewModel['checkboxes']['visible'][1]['checked'] );
+		$this->assertFalse( $viewModel['checkboxes']['visible'][2]['checked'] );
+		$this->assertFalse( $viewModel['checkboxes']['visible'][3]['checked'] );
+		$this->assertFalse( $viewModel['checkboxes']['visible'][4]['checked'] );
+
+		$this->assertFalse( $viewModel['checkboxes']['collapsed'][0]['checked'] );
+		$this->assertFalse( $viewModel['checkboxes']['collapsed'][1]['checked'] );
```
JeroenDeDauw, line 152: I see many existing tests got new entries and no new ones got added. There likely is an issue here. @malberts can you check?

src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -98,19 +98,33 @@ private function buildCheckboxesViewModel( FacetConfig $config, PropertyConstrai
...
+			} else {
+				$collapsedCheckboxes[] = $checkbox;
+			}
 		}
 
-		return $checkboxes;
+		return [
+			'visible' => $visibleCheckboxes,
+			'collapsed' => $collapsedCheckboxes,
+			'showMore' => count( $collapsedCheckboxes ) > 0,
```
JeroenDeDauw, line 124: Looks like you can return the method to its original form (returning just $checkboxes) and then have the caller (perhaps a new method) do

```php
return [
'visible' => array_slice_or_something_like_that(),
'collapsed' => array_slice_or_something_like_that(),
'showMore' => count( $checkboxes ) > 5,
];
```

JeroenDeDauw: That separates concerns better

Also makes so existing tests can stay focused on building of the checkbox view model, without having to care about the new logic.

tests/phpunit/Presentation/ListFacetHtmlBuilderTest.php
malberts: Some thoughts on the `StubValueCounter` change:
* It causes unnecessary/irrelevant assertions in other tests.
* Rather create a new test double, maybe `SequenceValueCounter`, that takes a `$count` int as a constructor parameter, and then generates `$count` amount of `ValueCount` in `countValues()`.
* Or else add that parameter in `StubValueCounter` and return only `$count` of the hardcoded values. But this adds some complication to the stub, so maybe it's better to keep it separate.

Regarding tests:
* add an explicit test for when the checkboxes count = boundary
* add an explicit test for when the count > boundary

malberts:
I merged master, but there seems to be an issue with other recent JS changes (likely https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/162). When clicking the "Show more" button it triggers a page reload.

JeroenDeDauw:
Eh, I just realized the bug is not fixed yet

Issue #166, opened 2025-01-29
Fix OR handling


malberts:
After https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/158

`QueryStringParser` defaults to AND when only one value is specified. This causes problems on the UI when the user tries to select OR mode, or when only one value is selected.

----

#### When changing to OR:

https://github.com/user-attachments/assets/5b7fff3f-90dc-456b-8ffa-76f45bdf6fe6

#### When toggling is enabled and OR is default:

https://github.com/user-attachments/assets/1b6872b7-ae40-4939-9d06-11813ecdb65f

----

#### When toggling is disabled and OR is default:

https://github.com/user-attachments/assets/c5204f9d-7de9-4bf1-9448-3a672c9a8baf

When manually changing the query to an OR query the UI works:

https://github.com/user-attachments/assets/2b7ae1f0-7cb7-4c2a-bc44-87a8980eea2a

----

The solution will be either:
* fix parsing to handle OR with single value
  * since additional query string parameters for the Search page will be non-standard, we might have to use a "hack" where we always include a "|", even if there is just one value
* update/set UI from client-side state
  * although the parsed query will then incorrectly indicate it is an AND facet when we show an OR

alistair3149:
Would it be favorable if we use a query such as `haswbfacet:P1460=P1|` to notify the backend that it is in OR mode?

malberts:
> Would it be favorable if we use a query such as haswbfacet:P1460=P1| to notify the backend that it is in OR mode?

I was just about to update the issue with that same suggestion. Yes, I think that will be an acceptible "hack". That will allow us to avoid adding additional URL query parameters, additional search query parameters, and additional client-side logic.

malberts:
Right, that case won't be solved by the above solution. Maybe we can do this just not reload the page (just toggle it). Then when a checkbox is selected it will pick up the selected OR.

This would not be shareable, but I think that is less of a concern here, since it is an empty query segment anyway.

JeroenDeDauw:
Agree sharability of AND/OR selection state when nothing is selected in a list facet is unimportant.

Being able to toggle to AND or OR while a list facet is empty is important. Not reloading the page when going that toggle while the list is empty seems like an excellent solution, and even enhances UX a tiny bit.

Issue #88, opened 2025-01-10
Implement search query quote parsing


malberts:
Follow-up to https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/31

Quote parsing is not implemented yet.

The skipped test added in https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/70 expected quotes around each OR value, to allow for multiple quoted values:
`haswbfacet:P42="foo bar"|second|"third value"`

However, the quote parsing that returns `$quotedValue` in `HasWbFacetFeature::doApply()` (via `SimpleKeywordFeature`) seems to expect quotes for the entire keyword value:
`haswbfacet:"P42=foo bar|second|third value"`

malberts:
If we make `HasWbFacetFeature` handle quotes as per the example test case, then our keyword would behave slightly differently compared to `haswbstatement`. But that might not be an issue, since or multi-value handling is already different (`P42=foo|bar|baz` vs `P42=foo|P42=bar|P42=baz`).

JeroenDeDauw:
I do not have a product-level preference for the quotation form, both seem viable enough, and we generate the search string via the UI anyway, so it's not central to the UX. So whatever is easier to implement, or up to the implementer

PR #167, opened 2025-01-29
Handle empty OR values

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/166

* In general: removes empty values: `haswbfacet:p42=foo||bar`
* Specific: allows a single OR value to be correctly parsed as an OR value: `haswbfacet:P42=Q100|`

### DIFF
--- a/src/Application/QueryStringParser.php
+++ b/src/Application/QueryStringParser.php
@@ -97,7 +97,7 @@ class QueryStringParser {
 
 		if ( $operator === '=' ) {
 			if ( str_contains( $value, '|' ) ) {
-				return $propertyConstraints->withOrValues( ...explode( '|', $value ) );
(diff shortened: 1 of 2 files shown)


src/Application/QueryStringParser.php
```diff
@@ -159,4 +159,13 @@ private function findOperator( string $constraint ): string {
 		return '';
 	}
 
+	/**
+	 * @return string[]
+	 */
+	private function getNonEmptyOrValues( string $value ): array {
+		return array_filter(
+			explode( '|', $value ),
+			fn( string $orValue ): bool => $orValue !== ''
```
malberts, line 168: Not handling whitespace, since that depends on quote handling (#88), otherwise the search engine will just consider it a separate token.

PR #169, opened 2025-01-31
Refactor list facet query generation logic

alistair3149:
Key changes:
- Move finding selected values from DOM from `getListFacetQuerySegments` to `getListFacetSelectedValues`
- Rename `checkedValues` to `selectedValues`
- Only pass `selectedValues` to `getListFacetQuerySegments` instead of the `facet` element

### DIFF
--- a/resources/ext.wikibase.facetedsearch.js
+++ b/resources/ext.wikibase.facetedsearch.js
@@ -84,8 +84,12 @@ function onInstancesClick( event, instanceId ) {
  * @param {?string} mode
  */
 function onListFacetInput( facet, propertyId, mode ) {
+	const selectedValues = getListFacetSelectedValues( facet );
+	if ( selectedValues.length === 0 ) {
+		return;
+	}
 	mode = mode || getListFacetQueryMode( facet );
-	const newQueries = getListFacetQuerySegments( facet, propertyId, mode );
+	const newQueries = getListFacetQuerySegments( selectedValues, propertyId, mode );
 	submitSearchForm( buildQueryString( specialSearchInput.value, newQueries, propertyId ) );
 }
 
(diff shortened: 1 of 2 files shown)


resources/ext.wikibase.facetedsearch.js
```diff
@@ -153,35 +157,41 @@ function updateErrorState( input ) {
 }
 
 /**
- * Extracts the queries from a list facet element.
+ * Extracts the selected facet items from a list facet element.
```
malberts, line 160: If we need to comment a function, then keeping the language consistent is important. The function refers to "facet values", whereas the comment refers to "facet items".

tests/jest/ext.wikibase.facetedsearch.test.js
```diff
@@ -38,37 +38,43 @@ describe( 'getListFacetQueryMode', () => {
...
+
+	test( 'Query segements for list facet with multiple selected values in AND mode', () => {
+		expect( actual.getListFacetQuerySegments( [ 'Q2', 'Q3' ], 'P1', 'AND' ) )
+			.toEqual( [ 'haswbfacet:P1=Q2', 'haswbfacet:P1=Q3' ] );
+	} );
 
-		expect( actual.getListFacetQuerySegments( document.body, 'P1', 'OR' ) )
+	test( 'Query segements for list facet with multiple selected values in OR mode', () => {
+		expect( actual.getListFacetQuerySegments( [ 'Q2', 'Q3' ], 'P1', 'OR' ) )
 			.toEqual( [ 'haswbfacet:P1=Q2|Q3' ] );
```
malberts, line 78: We should also have test cases for single selected value. This is going to be important for https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/166#issuecomment-2622295218.

resources/ext.wikibase.facetedsearch.js
```diff
@@ -84,8 +84,12 @@ function onInstancesClick( event, instanceId ) {
  * @param {?string} mode
  */
 function onListFacetInput( facet, propertyId, mode ) {
+	const selectedValues = getListFacetSelectedValues( facet );
+	if ( selectedValues.length === 0 ) {
+		return;
+	}
```
malberts, line 90: @alistair3149 I realise the error here now (https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/172). When deselecting everything, this function will return and not update the search.

Issue #134, opened 2025-01-19
Filter values and counts by current query


JeroenDeDauw:
AC:
* ElasticValueCounter returns ValueCounts for only results matching the current search query

Follow up to
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/23

Likely has overlap or interplay with
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/53

JeroenDeDauw:
@malberts is this done?

malberts:
@JeroenDeDauw No. The query generation for the search is separate from this. We still need to find a way to get the final query built by CirrusSearch and pass it into `ElasticValueCounter`.

PR #175, opened 2025-02-07
Inject current query

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/134

### DIFF
--- a/src/EntryPoints/WikibaseFacetedSearchHooks.php
+++ b/src/EntryPoints/WikibaseFacetedSearchHooks.php
@@ -6,8 +6,11 @@ namespace ProfessionalWiki\WikibaseFacetedSearch\EntryPoints;
 
 use CirrusSearch\CirrusSearch;
 use CirrusSearch\Query\KeywordFeature;
+use CirrusSearch\Search\CirrusSearchResultSet;
 use CirrusSearch\SearchConfig;
+use Elastica\Query\AbstractQuery;
 use HtmlArmor;
+use ISearchResultSet;
 use MediaWiki\Content\ContentHandler;
 use MediaWiki\EditPage\EditPage;
 use MediaWiki\Html\Html;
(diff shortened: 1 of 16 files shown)


src/EntryPoints/WikibaseFacetedSearchHooks.php
```diff
@@ -67,18 +71,39 @@ public static function onSpecialSearchResultsPrepend(
...
 			WikibaseFacetedSearchExtension::getInstance()->getSidebarHtmlBuilder( $specialSearch->getLanguage() )->createHtml(
-				searchQuery: $term
+				searchQuery: $term,
+				currentQuery: self::getCurrentQuery()
 			)
 		);
 	}
 
+	private static function getCurrentQuery(): AbstractQuery {
+		return $GLOBALS[WikibaseFacetedSearchExtension::QUERY_GLOBAL] ?? new MatchAll();
```
JeroenDeDauw, line 104: In which cases does the global not get set?

src/EntryPoints/WikibaseFacetedSearchHooks.php
```diff
@@ -67,18 +71,39 @@ public static function onSpecialSearchResultsPrepend(
...
+	public static function onSpecialSearchResults(
+		string $term,
+		?ISearchResultSet $titleMatches,
+		?ISearchResultSet $textMatches
+	): void {
+		if ( !( $textMatches instanceof CirrusSearchResultSet ) ) {
+			return;
+		}
+
+		self::setCurrentQuery( $textMatches->getElasticaResultSet()->getQuery()->getQuery() );
```
JeroenDeDauw, line 83: This is simpler:

```php
if ( $textMatches instanceof CirrusSearchResultSet ) {
    self::setCurrentQuery( $textMatches->getElasticaResultSet()->getQuery()->getQuery() );
}
```

src/EntryPoints/WikibaseFacetedSearchHooks.php
malberts: I removed it. This was more of a theoretical fallback, although I am not sure what would cause it. At least in our case, if any tab was selected that should always casue at least some query to exist.

PR #179, opened 2025-02-11
Handle empty selected values

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/166

Query keywords:
* `haswbfacet:P42=` means `AND` without affecting search
* `haswbfacet:P42=|` measn `OR` without affecting search
* `haswbfacet:P42=9000` means `AND` (same as before)
* `haswbfacet:P42=9000|` means `OR` (impact on query is same as `AND`)
* `haswbfacet:P42=9000|9001` means `OR` (same as before)

Implementation:
* query parser does not remove empty values
* caller of query parser handles empty values vs single empty string vs having values

https://github.com/user-attachments/assets/b06e36db-f1ce-4c8f-89c9-5472ca69edb2

### DIFF
--- a/src/Application/QueryStringParser.php
+++ b/src/Application/QueryStringParser.php
@@ -97,7 +97,7 @@ class QueryStringParser {
 
 		if ( $operator === '=' ) {
 			if ( str_contains( $value, '|' ) ) {
-				return $propertyConstraints->withOrValues( ...$this->getNonEmptyOrValues( $value ) );
+				return $propertyConstraints->withOrValues( ...explode( '|', $value ) );
 			}
 
 			return $propertyConstraints->withAdditionalAndValue( $value );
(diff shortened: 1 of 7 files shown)


src/Application/QueryStringParser.php
malberts, line 1: I considered making the parser remove all empty values and then setting an additional flag on `PropertyConstraints` to indicate that it should be considered OR or AND. But it wasn't immediately obvious how to handle that via the current constructor form. Since it allows having OR and AND values, the constructor would have to take an additional `isOr` parameter. And then either way, the `isOr` would have to be set in `withOrValues()`, or with an additional method call when adding OR values.

tests/phpunit/Application/QueryStringParserTest.php
```diff
@@ -275,4 +275,28 @@ public function testParsesSingleItemTypeWithAndValues(): void {
 		);
 	}
 
+	public function testHasSingleEmptyValueIfAndValueIsEmpty(): void {
+		$parser = $this->newQueryStringParser();
+		$query = $parser->parse( 'haswbfacet:P42=' );
+
+		$constraints = $query->getConstraintsForProperty( new NumericPropertyId( 'P42' ) );
+
+		$this->assertEquals(
```
JeroenDeDauw, line 284: These could be assertSame

Issue #181, opened 2025-02-11
Fix OR facet list values or aggregation query


malberts:
There might be a possible issue with the OR filter:

https://github.com/user-attachments/assets/143548df-0c51-4f7e-b85c-586c8fe7d15e

In some cases it filters the available values as if it was an AND query, and in other cases it displays all the values.

malberts:
We'll need to remove the OR query segment in the current query when counting the values for the OR filter.

If the current query is `haswbfacet:P598=Q57230|Q57229`, then it's going to return only additional occupations on items where `P598=Q57230` or `P598=Q57229`.

PR #183, opened 2025-02-11
Filter current query

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/181

https://github.com/user-attachments/assets/d4db1647-fce4-4159-a633-70f78d635ab9

### DIFF
--- a/tests/phpunit/Application/ElasticQueryFilterTest.php
+++ b/tests/phpunit/Application/ElasticQueryFilterTest.php
@@ -1,0 +1,224 @@
+<?php
+
+declare( strict_types = 1 );
+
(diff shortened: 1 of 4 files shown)


tests/phpunit/Application/ElasticQueryFilterTest.php
malberts, line 1: This file can be cleaned and improved, but I'm unsure of the value here. This does technically test that a query of a specific structure gets filtered, but I cannot test that these queries are the ones generated by the search, and therefore that a search's query will actually get filtered.

(However, I did build these simplified query objects here by looking at the final query object created by the search.)

Issue #176, opened 2025-02-08
Hide list facet when no values available


malberts:
The user cannot select anything, but can still collapse the filter and toggle AND/OR. This makes it look half broken.

Hiding empty filters seems to be the standard behavior for e-commerce facet filters anyway.

![Image](https://github.com/user-attachments/assets/03b98458-1fb5-4665-8dfe-4be07c39e8f2)

alistair3149:
Handling required for MVP is added in #185
The edge case in #180 will be implemented in a follow up PR.

PR #185, opened 2025-02-12
Add empty list facet handling

alistair3149:
https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/176

Key changes:
- Modify `ListFacetHtmlBuilder` to return empty string for facets with no values
- Update `SidebarHtmlBuilder` to hide facets when `facetHtml` is empty
- Adjust test cases to reflect new behavior

### DIFF
--- a/tests/phpunit/Presentation/ListFacetHtmlBuilderTest.php
+++ b/tests/phpunit/Presentation/ListFacetHtmlBuilderTest.php
@@ -4,7 +4,6 @@ declare( strict_types = 1 );
 
 namespace ProfessionalWiki\WikibaseFacetedSearch\Tests\Presentation;
 
-use Elastica\Query\MatchAll;
 use MediaWiki\MediaWikiServices;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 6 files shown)


tests/phpunit/Presentation/ListFacetHtmlBuilderTest.php
```diff
@@ -26,30 +25,6 @@ class ListFacetHtmlBuilderTest extends TestCase {
 
 	private const FACET_PROPERTY_ID = 'P42';
 
-	public function testRendersTemplate(): void {
```
malberts, line 29: Why did you remove this test? It still passes with the code change in this PR. Specifically, the StubValueCounter used here still returns items, so the facet should still render.

(The usefulness of this test is a different question, though.)

alistair3149: It was removed in an earlier iteration of the PR because it didn't pass, and that the test wasn't useful when we already checks the value of the checkbox in other tests.

Should the test be added back? If so, I'll do it in a follow up PR.

Issue #59, opened 2025-01-02
Trigger Page reindex when Item is updated


malberts:
Follow-up to:
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/56
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/51

If `Config::$linkTargetSitelinkSiteId` is not set, the facet values are indexed on the Item in Elasticsearch. When an Item is updated, it will get reindexed.

If `$linkTargetSitelinkSiteId` is set, then the values are indexed on the Page instead of the Item. When a Page is updated, it will get the statement values and apply them. However, when the Item is updated, it will not trigger a reindex, since the search data hook fires on the Item, not the page. Instead, a reindex should be triggered/scheduled for the sitelinked Page when the Item is updated.

AC:
* [ ] Page is reindexed when the sitelinked Item is updated

PR #188, opened 2025-02-17
Reindex item page

malberts:
For #59 

This creates a null edit on the linked page. Indexing of the page is then handled by MediaWiki (e.g. $wgJobRunRate or runJobs.php).

### DIFF
--- a/src/Persistence/SiteLinkItemPageUpdater.php
+++ b/src/Persistence/SiteLinkItemPageUpdater.php
@@ -1,0 +1,36 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Persistence;
+
+use MediaWiki\Page\WikiPageFactory;
+use MediaWiki\Title\Title;
+use MediaWiki\User\UserIdentity;
+use ProfessionalWiki\WikibaseFacetedSearch\Application\ItemPageUpdater;
+use Wikibase\DataModel\Entity\Item;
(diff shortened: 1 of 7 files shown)


src/Persistence/SiteLinkItemPageUpdater.php
malberts, line 1: I'm not immediately sure how to test this, besides checking if mocked methods get called. Otherwise I need to find some other side-effect of the null edit.

JeroenDeDauw: You did a manual test, right?

src/Persistence/SiteLinkItemPageUpdater.php
```diff
@@ -0,0 +1,36 @@
...
+		if ( $item->hasLinkToSite( $this->linkTargetSitelinkSiteId ) ) {
+			$this->getSiteLinkedPage( $item )
+				->newPageUpdater( $user )
+				->updateRevision();
+		}
+	}
+
+	private function getSiteLinkedPage( Item $item ): WikiPage {
+		return $this->pageFactory->newFromTitle(
+			Title::newFromText( $item->getSiteLink( $this->linkTargetSitelinkSiteId )->getPageName() )
```
JeroenDeDauw, line 32: `Title::newFromText` can return `null`, and this is not purely theoretical, as there is no guarantee that the value in the sitelink represents a valid page name on the current wiki.

Issue #177, opened 2025-02-08
Move selected facet filter values to the top


malberts:
Facet values are currently sorted by count (implicitly via the Elasticsearch aggregation in ElasticValueCounter, not explicitly in the query or code).

The 2 selected checkboxes here should be at the top, even if an unselected checkbox has a higher count:
![Image](https://github.com/user-attachments/assets/eac26507-ff2b-444b-ba62-e3e2f0ea933d)

JeroenDeDauw:
I am dubious that moving the values is good UX. If as a user I select the first and the third value, I do not expect the third value to get swapped with the second one. I'd probably end up thinking I misclicked until I notice they got swapped.

malberts:
I was basing this on a pattern used on e-commerce sites. Although it does not seem to be consistent.

Amazon moves the checkboxes for some facets (or maybe just the first?):
![Image](https://github.com/user-attachments/assets/c6162dae-c6d6-4aa2-a2b0-1323938178a5)

eBay moves them:
![Image](https://github.com/user-attachments/assets/e361e944-4025-42a8-8a15-45a6afc11da0)

Also, the order is not guaranteed anyway. If you select (1) and (3), and the resulting query leads to the exclusion of (2), or a recalculation of counts, then it's going to move anyway.

JeroenDeDauw:
Sounds plausible

PR #194, opened 2025-02-19
Sort selected facet values to the top

alistair3149:
#177

Key changes:
* Sort checkboxes to prioritize checked items
* Add an early exit in `buildCheckboxesViewModel` when there are no facet values (Unsure about this since theoretically this should not happen with #185)
* Update tests to reflect new sorting behavior

image

### DIFF
--- a/src/Presentation/ListFacetHtmlBuilder.php
+++ b/src/Presentation/ListFacetHtmlBuilder.php
@@ -105,6 +105,14 @@ class ListFacetHtmlBuilder implements FacetHtmlBuilder {
 	}
 
 	private function buildCheckboxesViewModel( FacetConfig $config, PropertyConstraints $state, array $valueCounts ): array {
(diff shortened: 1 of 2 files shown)


src/Presentation/ListFacetHtmlBuilder.php
```diff
@@ -119,6 +127,8 @@ private function buildCheckboxesViewModel( FacetConfig $config, PropertyConstrai
 			$checkboxes[] = $this->buildCheckboxViewModel( $config, $valueCount, $selectedValues, $state->propertyId, $i );
 		}
 
+		usort( $checkboxes, fn( $a, $b ) => $b['checked'] <=> $a['checked'] );
```
malberts, line 130: Does this keep the current order of the checked items?

alistair3149: Yes it does

PR #197, opened 2025-02-21
Add configuration link to UI

alistair3149:
Key changes:
- Add settings button to topbar that links to the configuration page, if the user has edit permissions
- Refactor topbar HTML to accommodate for the settings button

### DIFF
--- a/src/Presentation/TabsHtmlBuilder.php
+++ b/src/Presentation/TabsHtmlBuilder.php
@@ -5,11 +5,16 @@ declare( strict_types = 1 );
 namespace ProfessionalWiki\WikibaseFacetedSearch\Presentation;
 
 use MediaWiki\Html\TemplateParser;
+use MediaWiki\Title\Title;
 use ProfessionalWiki\WikibaseFacetedSearch\Application\Config;
 use ProfessionalWiki\WikibaseFacetedSearch\Application\ItemTypeLabelLookup;
 use ProfessionalWiki\WikibaseFacetedSearch\Application\Query;
 use ProfessionalWiki\WikibaseFacetedSearch\Application\QueryStringParser;
+use ProfessionalWiki\WikibaseFacetedSearch\WikibaseFacetedSearchExtension;
 use Wikibase\DataModel\Entity\ItemId;
+use MediaWiki\MediaWikiServices;
+use MediaWiki\Request\RequestContext;
+use MediaWiki\User\User;
 
 class TabsHtmlBuilder {
 
@@ -18,6 +23,7 @@ class TabsHtmlBuilder {
 		private readonly ItemTypeLabelLookup $itemTypeLabelLookup,
 		private readonly TemplateParser $templateParser,
 		private readonly QueryStringParser $queryStringParser,
+		private readonly User $user,
 	) {
 	}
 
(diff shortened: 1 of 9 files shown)


tests/phpunit/Presentation/TabsHtmlBuilderUnitTest.php
```diff
@@ -157,4 +190,68 @@ public function testTabsViewModelSelectsCurrentItemType(): void {
...
+		"Q5976445": {
+			"facets": {
+				"P592": {
+					"type": "list"
+				}
+			}
+		}
+	}
+}
+JSON );
```
JeroenDeDauw, line 207: Why are you providing the JSON? Can't we just call `newConfig()`, or am I missing a relevant detail?

tests/phpunit/Presentation/TabsHtmlBuilderUnitTest.php
```diff
@@ -157,4 +190,68 @@ public function testTabsViewModelSelectsCurrentItemType(): void {
...
+	"configPerItemType": {
+		"Q5976445": {
+			"facets": {
+				"P592": {
+					"type": "list"
+				}
+			}
+		}
+	}
+}
```
JeroenDeDauw, line 240: Same here

tests/phpunit/Presentation/TabsHtmlBuilderUnitTest.php
JeroenDeDauw: Actually, why the changes to the config code in this test class at all? Didn't the old `$config ?? new Config(),` work?

src/Presentation/TabsHtmlBuilder.php
```diff
@@ -47,6 +51,22 @@ private function parseQuery( string $searchQuery ): Query {
...
 	}
 
+	private function buildSettingsViewModel(): array {
+		$title = Title::newFromText( WikibaseFacetedSearchExtension::CONFIG_PAGE_TITLE, NS_MEDIAWIKI );
+
+		if (
+			!$title instanceof Title
+			|| !$this->user->isAllowed( 'editsitejson' )
+		) {
+			return [];
```
JeroenDeDauw, line 61: The logic here is not quite right

we need to check (wg)`WikibaseFacetedSearchEnableInWikiConfig`. Only show the link if it is `true`

we should not care about if the page already exists or not. If it does not exist, all the more reason for the people to go there and configure things

Lastly, I suspect there is a way to do something like `user->canEdit( $title )`. That is safer, since it checks all the relevant permissions, and avoids needing to know what those are in the `TabsHtmlBuilder`. That `canEdit` method does not exist and I cannot readily find the right one. It might well not be on the `User` class. I suspect we used it _somewhere_ a few months ago.

JeroenDeDauw:
That UI looks really good

tests/phpunit/Presentation/TabsHtmlBuilderUnitTest.php
alistair3149: Mainly to provide a bare minimum `Config` that does not throw an exception.

`new Config()` does not work because the tests will throw `RuntimeException: No instance of ID configured` when there are no `$config` provided. We did not encounter it before because every test cases already defined its own `$config`, so `new Config()` was never triggered.

Issue #196, opened 2025-02-21
Add configuration link to UI for admins


JeroenDeDauw:
Somewhere on the search page, when facets are shown and the user is allowed to edit the config page, show something that makes clear they can configure the UI, including a way to get to the config page.

Conceptually identical to what we did in Wikibase Export, where the highlighted text is only shown to people that can edit the config page:

![Image](https://github.com/user-attachments/assets/39432b5c-2854-4511-a7b2-8bf05275e0b2)

PR #202, opened 2025-02-24
Follow up to #196 

alistair3149:
Follow up to #196

Key changes
- Add `ConfigAuthorizer` class for checking if the user is authorized to edit the configuration page
- Add an early exit if the wiki does not allow configurating on the wiki
- Revert the change to the null value of the config in the test


tests/phpunit/Presentation/TabsHtmlBuilderUnitTest.php
```diff
@@ -211,7 +194,37 @@ public function testSettingsViewModelContainsValuesWhenUserCanEditConfigurations
...
+		"Q5976445": {
+			"facets": {
+				"P592": {
+					"type": "list"
+				}
+			}
+		}
+	}
+}
+JSON );
```
JeroenDeDauw, line 220: Continuing chat from https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/197#discussion_r1968140892

If this config JSON is here just to avoid the exception, then it is a detail that distracts from what this test is actually doing. Better to hide it behind a method like `$this->newValidConfig()`

JeroenDeDauw: Actually we already have such a function in the form of `Valid::config();`

src/Presentation/TabsHtmlBuilder.php
```diff
@@ -52,12 +55,13 @@ private function parseQuery( string $searchQuery ): Query {
...
+			return [];
+		}
+
 		$title = Title::newFromText( WikibaseFacetedSearchExtension::CONFIG_PAGE_TITLE, NS_MEDIAWIKI );
 
-		if (
-			!$title instanceof Title
-			|| !$this->user->isAllowed( 'editsitejson' )
-		) {
+		if ( !$title instanceof Title || !$this->permissionManager->userCan( 'edit', $this->user, $title ) ) {
```
JeroenDeDauw, line 64: The above nicely fits into a `ConfigurationAuthorizer` service which is injected here and itself takes 
* bool $enableWikiConfig,
* PermissionManager $permissionManager,
* User $user
* TitleFactory or whatever it is called, to avoid `Title::newFromText`

Adding multiple new dependencies to an existing service for one reason is a sign that _suggests_ a new service is called for

JeroenDeDauw: Plus, having 7 fields is also slightly suspect. It can happen, but most of the time, classes in a well designed codebase will have less. 7 fields is similar to 3 or 4 function parameters

src/Presentation/TabsHtmlBuilder.php
```diff
@@ -55,13 +53,9 @@ private function parseQuery( string $searchQuery ): Query {
 	}
 
 	private function buildSettingsViewModel(): array {
-		if ( !$this->enableWikiConfig ) {
-			return [];
-		}
-
-		$title = Title::newFromText( WikibaseFacetedSearchExtension::CONFIG_PAGE_TITLE, NS_MEDIAWIKI );
+		$title = $this->titleFactory->newFromText( WikibaseFacetedSearchExtension::CONFIG_PAGE_TITLE, NS_MEDIAWIKI );
```
alistair3149, line 56: I am not sure about having `TitleFactory` in this class, but we need `Title` to get the URL and the `PageIdentity` object needed for checking permissions.

Apart from that, we used `Title::newFromText` in multiple places to get the `Title` object for the configuration page. Should that be turned into a class?

src/Application/ConfigAuthorizer.php
```diff
@@ -0,0 +1,26 @@
...
+		private readonly User $user
+	) {
+	}
+
+	public function isAuthorized( ProperPageIdentity $page ): bool {
+		if ( !$this->enableWikiConfig ) {
+			return false;
+		}
+
+		return $this->user->probablyCan( 'edit', $page );
```
alistair3149, line 23: `probablyCan()` is used for performance reason to avoid expensive DB query

src/Application/ConfigAuthorizer.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
+
+use MediaWiki\Page\ProperPageIdentity;
+use MediaWiki\User\User;
+
+class ConfigAuthorizer {
+
+	public function __construct(
+		private readonly bool $enableWikiConfig,
```
malberts, line 13: Minor nitpick: `$enableWikiConfig` sounds like the name of a method/action. `$wikiConfigIsEnabled` sounds more like a boolean.

src/Presentation/TabsHtmlBuilder.php
```diff
@@ -6,22 +6,24 @@
...
 
 	public function __construct(
 		private readonly Config $config,
+		private readonly ConfigAuthorizer $configAuthorizer,
 		private readonly ItemTypeLabelLookup $itemTypeLabelLookup,
 		private readonly TemplateParser $templateParser,
-		private readonly QueryStringParser $queryStringParser,
-		private readonly User $user,
+		private readonly TitleFactory $titleFactory,
+		private readonly QueryStringParser $queryStringParser
```
malberts, line 26: Did you arrange these alphabetically? Usually we don't do that, and instead just add new collaborators at the end.

tests/phpunit/Presentation/TabsHtmlBuilderUnitTest.php
```diff
@@ -39,50 +41,31 @@ public function testTabsViewModelContainsItemTypeProperty(): void {
...
-	private function newUser( bool $canEditConfig = false ): User {
-		$user = $this->createMock( User::class );
-		$user->method( 'isAllowed' )
-			->with( 'editsitejson' )
+	private function newConfigAuthorizer( bool $canEditConfig = false ): ConfigAuthorizer {
+		$configAuthorizer = $this->createMock( ConfigAuthorizer::class );
+		$configAuthorizer->method( 'isAuthorized' )
 			->willReturn( $canEditConfig );
-		return $user;
+		return $configAuthorizer;
```
malberts, line 64: We generally try to avoid mocks. When you need a test double, you can change `ConfigAuthorizer` to an interface, and the current `ConfigAuthorizer` becomes something like `UserBasedConfigAuthorizer`. Then you create a test double like `SucceedingConfigAuthorizer` that always returns true.

src/Presentation/TabsHtmlBuilder.php
JeroenDeDauw: Using `TitleFactory` instead of `Title::newFromText` isn't hugely important if we do not make use of the polymorphic seam, such as via a test double. Then again, it is fairly easy to default to the OO approach of injecting `TitleFactory`

src/Application/ConfigAuthorizer.php
```diff
@@ -0,0 +1,26 @@
...
+		private readonly User $user
+	) {
+	}
+
+	public function isAuthorized( ProperPageIdentity $page ): bool {
```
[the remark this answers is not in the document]
alistair3149: I use `ProperPageIdentity` because it is the output of `$title->toPageIdentity()`. As you mentioned, `probablyCan()` only requires `PageIdentity`.

> avoid calling $title->toPageIdentity() later.

What do you mean by that?

tests/phpunit/Presentation/TabsHtmlBuilderUnitTest.php
JeroenDeDauw: https://www.entropywins.wtf/blog/2016/05/13/5-ways-to-write-better-mocks/

Issue #198, opened 2025-02-22
Show links for item type tab translations


malberts:
On the config page `MediaWiki:WikibaseFacetedSearch`:
* create a new documentation section for item types
* show links to `MediaWiki:WikibaseFacetedSearch-item-type-Q42` for each of the item types defined in the config

PR #210, opened 2025-02-26
Add item type labels to config documentation

alistair3149:
Closes: #198

Key changes:
- Add item type label section
- Refactor config documentation into Mustache
- Fix minor typo

image

### DIFF
--- a/src/Presentation/ConfigEditPageTextBuilder.php
+++ b/src/Presentation/ConfigEditPageTextBuilder.php
@@ -5,113 +5,38 @@ declare( strict_types = 1 );
 namespace ProfessionalWiki\WikibaseFacetedSearch\Presentation;
 
 use MediaWiki\Context\IContextSource;
+use MediaWiki\Html\TemplateParser;
+use ProfessionalWiki\WikibaseFacetedSearch\Application\Config;
 
 class ConfigEditPageTextBuilder {
 
 	public function __construct(
 		private readonly IContextSource $context,
-		private readonly string $exampleConfigPath
+		private readonly string $exampleConfigPath,
+		private readonly TemplateParser $templateParser,
+		private readonly Config $config
 	) {
 	}
 
 	public function createTopHtml(): string {
-		return '<div id="wikibase-faceted-search-config-help-top">' .
-			$this->createDocumentationLink() .
-			'</div>';
-	}
-
-	private function createDocumentationLink(): string {
-		return '<p>'
-			. $this->context->msg( 'wikibase-faceted-search-config-help-documentation' )->parse()
-			. '</p>';
+		return $this->templateParser->processTemplate(
+			'ConfigEditPageTop',
+			[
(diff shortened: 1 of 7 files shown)


JeroenDeDauw:
Product thoughts on `item type labels` section:

image

* Change column header from `ID` to `Item`
* Show the Item's Wikibase Label in brackets next to the ID
* Link the entire text to the page (wiki/Item:Q123)
* The column header `Label` is confusing, since it is not the Wikibase Label. Same for the section header and the description: we should not call it "label". Something like "Tab header text" or "Tab name" solves the confusion issue by being distinct, and also makes it obvious what the thing is for.

alistair3149:
> Having the table with edit links is great. Putting it in the documentation might not be optimal. It's somewhat surprising to have to scroll through docs to find a UI for editing some specific config. Could move this UI up, though unsure if that works out better.
> 
> Things would be less weird if we had a reactive UI for the entire config (rather than editing the JSON directly), but that is a bunch of work to create.

As a stopgap solution, we can add an anchor link that links to the table to the top of the configuration page. We can revisit the reactive UI solution in the future

src/Presentation/ConfigEditPageTextBuilder.php
```diff
@@ -124,4 +49,31 @@ private function getExampleContents(): string {
 		return $example;
 	}
 
+	private function getItemTypesData(): array {
+		$itemTypes = $this->config->getItemTypes();
+		if ( $itemTypes === [] ) {
+			return [];
+		}
```
malberts, line 56: I think in cases like this it is not necessary to have an early exit. If the logic is based on an array loop, then the loop will do nothing and return an empty array anyway. The early exit suggests the empty array needs to be handled explicitly.

This can also be seen if the logic here was replaced by something like:
```php
return array_map(
    fn( $itemType ) => $this->getItemTypeData( $itemType ),
    $itemTypes
);
```

src/Presentation/ConfigEditPageTextBuilder.php
```diff
@@ -124,4 +49,31 @@ private function getExampleContents(): string {
...
+
+		$data = [];
+		foreach ( $itemTypes as $itemType ) {
+			$id = $itemType->getSerialization();
+			if ( !$id ) {
```
malberts, line 61: This check should be explicit. However, why do we need to check for this? The `ItemId` constructor already checks that the format is valid. So the `ItemId`s you get from the config are already valid.

malberts:
Regarding the table position: I also think we should move it directly under the form, above the documentation. The table is part of the configuration, not documentation, so should be grouped with the rest of the config.

alistair3149:
I have addressed all the issues in code review and also:
- Add i18n for all parts of the table
- Use `LinkRenderer` for proper handling for URLs

I need some refactor suggestions on breaking down the class further since there might be too much dependencies in the constructor.

JeroenDeDauw:
> I need some refactor suggestions on breaking down the class further since there might be too much dependencies in the constructor.

You could extract the code that builds the table. That has its own set of dependencies and does not need the template renderer. This also naturally exposes the view model it creates via a public method, making testing easier.

Issue #209, opened 2025-02-26
Support icons for item types


JeroenDeDauw:
![Image](https://github.com/user-attachments/assets/b83145a7-8ac1-495d-8b6a-5fabc1445562)

![Image](https://github.com/user-attachments/assets/9dd705de-7729-4702-bafe-0b3f5d517093)

alistair3149:
For the icon parameter, do we know what icon library that the client is using?

JeroenDeDauw:
No. What did you use in that screenshot? Did I understand correctly that you think using utf8 emoji is a bad idea?

alistair3149:
> What did you use in that screenshot?

Since it was only used as a mockup, it was a dirty HTML edit by dumping the whole SVG and adding some CSS...

> Did I understand correctly that you think using utf8 emoji is a bad idea

Emoji doesn't cover all the use cases, cannot be recolored, and the appearance differs depending on the font used.
It is more useful to use either an icon font or just render the icon directly. Since they are already using Chameleon as you mentioned, FontAwesome would probably be the way to go.

malberts:
I suppose allowing (some) HTML in `icon` to allow using different libraries is not worth the pain of making the HTML safe.

So perhaps a future solution is to have something like `iconProvider`, which can then be changed to Codex or whatever else. Of course, not needed for now.

PR #212, opened 2025-02-27
Add Font Awesome icon support to tab name

alistair3149:
Related: #209

Key changes:
- Add `IconBuilder` and `FontAwesomeIconBuilder` classes for creating icon HTML
- Add `icon` key to the values of `configPerItemType`. Currently, this is used to define the key for Font Awesome icons

Notes:
- `FontAwesomeIconBuilder` is aimed to support v6, but it should work with v5 as well. It is built that way because the Font Awesome implementation for MW is not great.
[Skin:Chameleon](https://github.com/ProfessionalWiki/chameleon) is on v5, and [Extension:FontAwesome](https://github.com/ProfessionalWiki/FontAwesome) is on v6 but it is still using the old style classes, both are not up-to-date with the latest practices. We should consider updating Extension:FontAwesome to drop the old classes.
- I didn't make Font Awesome an dependency, it is up to the wiki to load Font Awesome since it can be loaded in multiple ways. If the wiki loads Font Awesome through CSS or JS, the extension wouldn't be able to tell.
- The icon configuration works but I am not sure if this is the right approach to it. This might contribute to #107.

To-do:
- Update https://github.com/ProfessionalWiki/mediawiki-dev/tree/wbfs with the new config

image

### DIFF
--- a/src/Application/Config.php
+++ b/src/Application/Config.php
@@ -14,7 +14,7 @@ class Config {
 		public readonly ?string $sitelinkSiteId = null,
 		private readonly ?PropertyId $itemTypeProperty = null,
 		private readonly ?FacetConfigList $facets = null,
-		private readonly ?array $icons = null
+		private readonly ?array $icons = null,
 	) {
 	}
 
@@ -23,7 +23,7 @@ class Config {
 			$config->sitelinkSiteId ?? $this->sitelinkSiteId,
 			$config->itemTypeProperty ?? $this->itemTypeProperty,
 			$config->facets ?? $this->facets,
-			$config->icons ?? $this->icons
+			$config->icons ?? $this->icons,
 		);
 	}
 
--- a/src/Application/Config.php
+++ b/src/Application/Config.php
@@ -13,7 +13,8 @@ class Config {
 	public function __construct(
 		public readonly ?string $sitelinkSiteId = null,
 		private readonly ?PropertyId $itemTypeProperty = null,
-		private readonly ?FacetConfigList $facets = null
+		private readonly ?FacetConfigList $facets = null,
+		private readonly ?array $icons = null
 	) {
 	}
 
@@ -21,7 +22,8 @@ class Config {
 		return new Config(
 			$config->sitelinkSiteId ?? $this->sitelinkSiteId,
 			$config->itemTypeProperty ?? $this->itemTypeProperty,
-			$config->facets ?? $this->facets
+			$config->facets ?? $this->facets,
+			$config->icons ?? $this->icons
 		);
 	}
 
@@ -35,6 +37,10 @@ class Config {
 
 	public function getFacets(): FacetConfigList {
 		return $this->facets ?? new FacetConfigList();
+	}
+
+	public function getIconForItemType( ItemId $itemType ): ?string {
+		return $this->icons[$itemType->getSerialization()] ?? null;
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


src/Application/Config.php
```diff
@@ -13,15 +13,17 @@ class Config {
...
 	) {
 	}
 
 	public function combine( Config $config ): self {
 		return new Config(
 			$config->sitelinkSiteId ?? $this->sitelinkSiteId,
 			$config->itemTypeProperty ?? $this->itemTypeProperty,
-			$config->facets ?? $this->facets
+			$config->facets ?? $this->facets,
+			$config->icons ?? $this->icons
```
JeroenDeDauw, line 26: Example of a line change that could have been avoided if the tailing comma was already there. Still, this only matters minutely

src/Persistence/ConfigDeserializer.php
```diff
@@ -35,10 +35,16 @@ public function deserialize( string $configJson ): Config {
 	 * @param array<string, mixed> $configArray
 	 */
 	private function newConfig( array $configArray ): Config {
+		// Need to stop phpstan from complaining about the type of $configPerItemType
+		$configPerItemType = isset( $configArray['configPerItemType'] ) && is_array( $configArray['configPerItemType'] )
+			? $configArray['configPerItemType']
+			: [];
+
```
JeroenDeDauw, line 42: I rather not add such code that is there only to make a tool happy, at least in cases where we can easily fix the issue on the tool side by changing config or adding the issue to the baseline file.

I like you added the comment

JeroenDeDauw: You can update the baseline file with `make stan-baseline` from within the container

src/Persistence/ConfigDeserializer.php
```diff
@@ -50,9 +56,24 @@ private function newPropertyId( ?string $propertyId ): ?PropertyId {
 		return new NumericPropertyId( $propertyId );
 	}
 
-	/**
-	 * @param array<string, array> $configPerItemType
-	 */
+	private function newIconsList( array $configPerItemType ): ?array {
+		if ( $configPerItemType === [] ) {
+			return null;
+		}
```
JeroenDeDauw, line 62: This check seems redundant, given you also check for empty array in the line with the return

src/Presentation/TabsHtmlBuilder.php
```diff
@@ -91,9 +88,31 @@ private function buildTabsViewModel( ?ItemId $selectedItemType ): array {
...
+			$tab['icon'] = $icon;
+		}
+
+		return $tab;
+	}
+
+	private function buildTabIcon( ItemId $itemType ): ?string {
+		$icon = $this->config->getIconForItemType( $itemType );
+
+		if ( $icon ) {
```
JeroenDeDauw, line 109: `!== null`

JeroenDeDauw: Or `=== null ) { return null }`

src/Presentation/FontAwesomeIconBuilder.php
```diff
@@ -0,0 +1,67 @@
...
+			"fa-$iconName"
+		];
+
+		if ( isset( self::STYLE_SHORT_FORMS[$style] ) ) {
+			$classes[] = self::STYLE_SHORT_FORMS[$style];
+		}
+
+		// Font Awesome omits the family class for the 'classic' style
+		if ( $family !== 'classic' ) {
+			$classes[] = "fa-$family";
```
JeroenDeDauw, line 59: I noticed the test class has only assertStringContainsString and no assertStringNotContainsString. This if statement does not seem covered

JeroenDeDauw:
I did not spot any serious issues. Needs rebase

PR #217, opened 2025-03-05
Reword Settings button

JeroenDeDauw:
New
image

Original
image

## Motivation

We use "configuration" everywhere, not settings. So "configuration" seems better. And since it is a button, "configure" seems better.

I wanted to switch to a wrench icon but could not find. Perhaps this edit one is also good

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -23,7 +23,7 @@
 	"wikibase-faceted-search-instance-type-all": "All",
 
 	"wikibase-faceted-search-filters": "Filters",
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
@alistair3149 do you fully agree? I wonder how this compares to other stuff

alistair3149:
> @alistair3149 do you fully agree? I wonder how this compares to other stuff

I don't have a strong preference for edit or settings. Both are semantically correct and fit into users' expectations.

Issue #191, opened 2025-02-17
Update docs on config page


JeroenDeDauw:
Update and complete the docs on `MediaWiki:WikibaseFacetedSearch`

alistair3149:
Blocked by potential config refactor.
Check again on Wednesday to see if we can proceed.

JeroenDeDauw:
There will be no config refactor

PR #218, opened 2025-03-06
Update documentation

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/191

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -59,3 +59,9 @@ parameters:
 			identifier: argument.type
 			count: 1
 			path: src/Persistence/Search/Query/ListFacetQueryBuilder.php
+
+		-
+			message: '#^Cannot call method getFullURL\(\) on MediaWiki\\Title\\Title\|null\.$#'
+			identifier: method.nonObject
+			count: 1
+			path: src/Presentation/ConfigEditPageTextBuilder.php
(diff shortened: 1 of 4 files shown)


src/Presentation/ConfigEditPageTextBuilder.php
```diff
@@ -45,7 +45,8 @@ public function createBottomHtml(): string {
 				'msg-wikibase-faceted-search-config-tab-name-missing' => $this->context->msg( 'wikibase-faceted-search-config-tab-name-missing' )->plain(),
 				'msg-wikibase-faceted-search-config-help-example' => $this->context->msg( 'wikibase-faceted-search-config-help-example' )->escaped(),
 				'exampleContents' => $this->getExampleContents(),
-				'array-itemTypes' => $this->getItemTypesData()
+				'array-itemTypes' => $this->getItemTypesData(),
+				'searchUrl' => $this->getSearchUrl(),
```
malberts, line 49: I did not use a message string here for the full link, because the rest of the documentation is not translatable. It would look strange to have individual sentences in another language while the rest of the documentation cannot be translated. If we decide to make the documentation translatable then this can go away.

phpstan-baseline.neon
```diff
@@ -59,3 +59,9 @@ parameters:
 			identifier: argument.type
 			count: 1
 			path: src/Persistence/Search/Query/ListFacetQueryBuilder.php
+
+		-
+			message: '#^Cannot call method getFullURL\(\) on MediaWiki\\Title\\Title\|null\.$#'
```
malberts, line 64: I ignored this for now, since "Special:Search" should exist. If there's some wiki-specific config that can make the page disappear, then we probably need to make the whole sentence dynamic or find the search page in a different way.

PR #219, opened 2025-03-06
Show documentation when viewing config page

malberts:
Without config on the wiki:
image

With config on the wiki:
image

### DIFF
--- a/src/WikibaseFacetedSearchExtension.php
+++ b/src/WikibaseFacetedSearchExtension.php
@@ -7,6 +7,7 @@ namespace ProfessionalWiki\WikibaseFacetedSearch;
 use CirrusSearch\CirrusSearch;
 use CirrusSearch\SearchConfig;
 use Elastica\Query\AbstractQuery;
+use MediaWiki\Context\IContextSource;
 use MediaWiki\Html\TemplateParser;
 use MediaWiki\Language\Language;
 use MediaWiki\Linker\LinkRendererFactory;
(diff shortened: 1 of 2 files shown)


src/WikibaseFacetedSearchExtension.php
```diff
@@ -374,4 +376,16 @@ public function newItemPageUpdater(): ItemPageUpdater {
 		);
 	}
 
+	public function newConfigEditPageTextBuilder( IContextSource $context ): ConfigEditPageTextBuilder {
+		return new ConfigEditPageTextBuilder(
```
JeroenDeDauw, line 380: This name is weird now

Better to make it be about what it builds, rather than about where what it builds is used

JeroenDeDauw:
The "Configure tab Headers" section is a bit odd on the view page. Especially since the viewers might not be able to "create" or "edit" the linked pages.

Could show it only on the edit page. Perhaps a good time to extract that tab header stuff from the `ConfigEditPageTextBuilder`

src/WikibaseFacetedSearchExtension.php
malberts: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/220

malberts:
>  "Configure tab Headers" section

Maybe just hide the create/edit link if the user does not have permission. It might still be useful to show the table without that.

JeroenDeDauw:
I dunno which of those approaches is better

alistair3149:
> > "Configure tab Headers" section
> 
> Maybe just hide the create/edit link if the user does not have permission. It might still be useful to show the table without that.

I agree that the table are still useful to have.
We can modify the link text to `view` if the user has no permission. In addition, maybe we can rename the `Configure tab Header` text to something else such as just `Tab header`

Issue #236, opened 2025-03-12
Documentation: extension depends on  'WikibaseCirrusSearch'


physikerwelt:
```php
./maintenance/run update --quick
PHP Fatal error:  Uncaught MediaWiki\Registration\ExtensionDependencyError: Wikibase Faceted Search requires WikibaseCirrusSearch to be installed.
 in /Users/physikerwelt/git/mw/includes/registration/ExtensionRegistry.php:477

PR #237, opened 2025-03-12
Remove WikibaseCirrusSearch dependency

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/236

WikibaseCirrusSearch was used early on when we depended on the `haswbstatement` keyword (and possibly something else). However, the code no longer depends on it, so it can be removed.


physikerwelt (approved):
LGTM

physikerwelt:
tested, works

PR #241, opened 2025-03-16
Use namespace

malberts:
For #231 

----

Elasticsearch data comparison to https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/231#issuecomment-2726535040

`namespace` contains a value, and `title` no longer includes the namespace

```bash
curl -X GET "http://elasticsearch:9200/_search?pretty" -H "Content-Type: application/json" -d '
{
  "query": {
    "term": {
      "_id": "47"
    }
  }
}'
```
```
...
        "_index" : "mediawiki_content_first",
        "_type" : "_doc",
        "_id" : "47",
        "_score" : 1.0,
        "_source" : {
          "version" : 47,
          "wiki" : "mediawiki",
          "page_id" : 47,
          "namespace" : 4202,
          "namespace_text" : "Person",
          "title" : "62906",
          "timestamp" : "2023-12-06T12:49:42Z",
          "create_timestamp" : "2023-12-06T12:49:42Z",
          "redirect" : [ ],
          "incoming_links" : 0,
          "wbfs_P592" : [
            "Q57196"
          ],
          "wbfs_P598" : [
            "Q57229",
            "Q62432",
            "Q59697",
            "Q61708"
          ],
          "wbfs_P593" : [
            "1979-10-14T00:00:00Z"
          ],
          "wbfs_P437" : [ ],
          "wbfs_P463" : [
            140312.0
          ],
          "wbfs_P1460" : [
            "Q5976445"
          ],
...
```

PR #242, opened 2025-03-16
Sitelink based statements lookup test

JeroenDeDauw:
Follow up to https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/241

### DIFF
--- a/tests/phpunit/Persistence/SitelinkBasedStatementsLookupTest.php
+++ b/tests/phpunit/Persistence/SitelinkBasedStatementsLookupTest.php
@@ -41,18 +41,18 @@ class SitelinkBasedStatementsLookupTest extends WikibaseFacetedSearchIntegration
 		);
 	}
 
-	public function testPageWithoutSitelinkReturnsNoStatements(): void {
-		$this->assertPageHasStatements( $this->createPage(), [] );
+	public function testReturnsNoStatementsForPageWithoutIncomingSitelink(): void {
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Ugh some stuff went missing from my commits

tests/phpunit/Persistence/SitelinkBasedStatementsLookupTest.php
```diff
@@ -41,18 +41,18 @@ protected function setUp(): void {
 		);
 	}
 
-	public function testPageWithoutSitelinkReturnsNoStatements(): void {
```
JeroenDeDauw, line 44: This name is misleading in two ways:
* The page is not returning anything, it's the lookup under test that is
* The page has no sitelinks. It's being referenced by a sitelink (or not in this case)

tests/phpunit/Persistence/SitelinkBasedStatementsLookupTest.php
```diff
@@ -41,18 +41,18 @@ protected function setUp(): void {
...
-		$this->assertPageHasStatements( $this->createPage(), [] );
```
JeroenDeDauw, line 45: Besides the same naming issue, this assertion method is hiding the Act part of the test. I didn't bother refactoring it now, but looks like it would have been better if this code remained inlined

Issue #246, opened 2025-03-18
Simplify PageItemLookup implementations


JeroenDeDauw:
We overcomplicated things with the PageItemLookup implementations. We have NullPageItemLookup and SitelinkPageItemLookup constructed via PageItemLookupFactory. These 3 can be combined into one implementation of PageItemLookup without any design downsides and some upsides.

PR #248, opened 2025-03-18
Simplify PageItemLookup implementations

alistair3149:
#246

Merged `NullPageItemLookup` and  `PageItemLookupFactory` into `SitelinkPageItemLookup`


tests/phpunit/Persistence/SitelinkItemPageLookupTest.php
```diff
@@ -29,6 +29,15 @@ protected function setUp(): void {
...
 	}
 
+	public function testReturnsNullWhenNoSitelinkSiteIdIsConfigured(): void {
+		$this->assertNull(
+			( new SitelinkPageItemLookup(
+				$this->sitelinkStore,
+				null
+			) )->getItemId( Title::newFromText( 'Foo' ) )
+		);
+	}
```
JeroenDeDauw, line 39: This test isn't great because there are multiple reasons for null to be returned. If you break the behavior under test, you'd still get null.

JeroenDeDauw: Then again, no need to start cleaning up this test class

Issue #254, opened 2025-03-25
Fix year range filtering


malberts:
A page like https://portal.mardi4nfdi.de/wiki/Publication:769271 has the following year data in Elasticsearch:
```
"wbfs_P28" : [
  "1957-00-00T00:00:00Z"
],
```

Current code creates a range:
https://github.com/ProfessionalWiki/WikibaseFacetedSearch/blob/09b7f20aa0aa98c2414fb1e2aeb17a674770e641/src/Persistence/Search/Query/RangeFacetQueryBuilder.php#L40-L48

For example, when filtering:
![Image](https://github.com/user-attachments/assets/eedaa2e8-96e3-4c0d-88e2-d637af12492c)

The query is going to be `1957-01-01 <= x <= 1957-12-31`, which presumably leaves `1957-00-00` out of bounds.

Possible solutions:
* change the lower bound to `-00-00`; or
* change lower bound to end of previous year (e.g. `1956-12-31`)
* index the value as a valid ISO 8601 date (e.g. `1957-01-01`)

PR #255, opened 2025-03-25
Add basic date precision handling

malberts:
For #254 

Turns out we have to index the partial date as a valid ISO 8601 date, otherwise the date range filter ignores it.

----

Search returns:
Screenshot_20250325_231000

For statement:
Screenshot_20250325_231048

With Elastic data:
```
"wbfs_P28" : [
  "1957-01-01T00:00:00Z"
],

```

### DIFF
--- a/src/Application/DataValueTranslator.php
+++ b/src/Application/DataValueTranslator.php
@@ -22,7 +22,8 @@ class DataValueTranslator {
 		}
 
 		if ( $dataValue instanceof TimeValue ) {
-			return ltrim( $dataValue->getTime(), '+-' );
+			// TODO: handle date precision explicitly.
+			return str_replace( '-00', '-01', ltrim( $dataValue->getTime(), '+-' ) );
 		}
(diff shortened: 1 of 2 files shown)


src/Application/DataValueTranslator.php
```diff
@@ -22,7 +22,8 @@ public function translate( DataValue $dataValue ): mixed {
 		}
 
 		if ( $dataValue instanceof TimeValue ) {
-			return ltrim( $dataValue->getTime(), '+-' );
+			// TODO: handle date precision explicitly.
+			return str_replace( '-00', '-01', ltrim( $dataValue->getTime(), '+-' ) );
```
malberts, line 26: This replaces the "missing" date parts for year/month/day precisions. Other precisions are not supported.

JeroenDeDauw:
This is the type of thing that is ideally also integration tested, because that is the level the bug occurred on

Issue #258, opened 2025-03-29
Incorrect handling of OR in list facet


JeroenDeDauw:
The first author has 37k results. If I select a classification with only 8k results, the author still has 37k, which is not possible (should be max 8k). This happens only if the classification facet is in OR mode.

https://github.com/user-attachments/assets/cc432dba-dd14-4059-b9c8-34989638bb3f

It's correct for the OR mode to not reduce result counts within the facet itself, but it should reduce them for the other facets, since those are combined with AND.

PR #259, opened 2025-03-30
Remove specific property when counting

malberts:
For #258 

Previous problem: It removed _all_ OR queries from the current query.
Solution: Remove the OR query only when getting counts for the specific facet. 

[Screencast_20250330_135332.webm](https://github.com/user-attachments/assets/bcc2d3cb-02e1-4167-9a20-43bc927dedd2)

### DIFF
--- a/src/Application/ValueCounter.php
+++ b/src/Application/ValueCounter.php
@@ -4,10 +4,8 @@ declare( strict_types = 1 );
 
 namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
 
-use Wikibase\DataModel\Entity\PropertyId;
-
 interface ValueCounter {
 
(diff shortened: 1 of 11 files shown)


src/Application/ValueCounter.php
```diff
@@ -4,10 +4,8 @@
 
 namespace ProfessionalWiki\WikibaseFacetedSearch\Application;
 
-use Wikibase\DataModel\Entity\PropertyId;
-
 interface ValueCounter {
 
-	public function countValues( PropertyId $property ): ValueCounts;
+	public function countValues( PropertyConstraints $constraints ): ValueCounts;
```
malberts, line 9: I guess this slightly changes the semantics of the method (i.e. counting the _state_, instead of a _property_), but PropertyConstraints contains PropertyId, so it seemed silly to include both.

Issue #119, opened 2025-01-15
Add "any value" option to list facets


JeroenDeDauw:
Add the ability to filter results by "any value" in list facets

AC:
* New UI element. Figure out sensible UI behavior (when to show / hide / disable / (de)select automatically)
* Act on existing `showAnyFilter` config in `ListFacetHtmlBuilder`
* Generates query segment in this format: `haswbfacet:P1`

Related
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/117

malberts:
Ignoring the final labels for now, here's a simple initial idea (including #117) that should involve only minimal backend changes.

UI:
![Image](https://github.com/user-attachments/assets/c9e92979-a137-4aed-b2f1-cce9c762c8a8)

"Any value" does _not_ show a count.
"No value" does _not_ show a count.

When "Any value" is checked, everything else gets unchecked.
When "No value" is checked, everything else gets unchecked.
When a specific value is checked, "Any value" and "No value" get unchecked.

"Any value" and "No value" are always shown (if `showAnyFilter` and `showNoneFilter` = true).
No hiding/disabling of checkboxes based on the current selection, making it easier for the user to switch.

Big caveat:
* the selected checkbox affects the current query, which in turn affects the available values and counts
* if "Any value" is checked, then all other specific values/counts will be shown, because "Any value" is a superset of any specific value
* if "No value" is checked, then all other specific values/counts will _not_ be shown, because the current query necessarily excludes any specific value
  * This should not be a blocker for this issue, but will require attention in #117

JeroenDeDauw:
All makes sense to me

Issue #117, opened 2025-01-15
Add "no value" option to list facets


JeroenDeDauw:
Add the ability to filter results by "no value" in list facets

AC:
* New UI element. Figure out sensible UI behavior (when to show / hide / disable / (de)select automatically)
* Act on existing `showNoneFilter` config in `ListFacetHtmlBuilder`
* Generates query segment in this format: `-haswbfacet:P1`

Related:
* https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/119

JeroenDeDauw:
Partial UI behavior idea: Should not be combinable with values using AND (because this does not make sense: "there should be no Author and the Author should be `Bob`")

malberts:
> Partial UI behavior idea: Should not be combinable with values using AND (because this does not make sense: "there should be no Author and the Author should be Bob")

This can be avoided by hiding it:
* no value selected -> show "no value" option
* any value selected -> hide "no value" option
* "no value" selected -> hide all other options

This hiding behavior should be consistent with how empty facet list values get hidden after selecting a value.

JeroenDeDauw:
> * any value selected -> hide "no value" option
> * "no value" selected -> hide all other options

@malberts you posted a video from some e-commerce site a while ago. IIRC one did not hide the options but automatically removed selection. So if you select "any value", the others would get deselected and remain visible, so that you can click "Bob (5)" at which point "any value" gets deseleted. This seems like better UX. If you hide the options, it becomes less obvious you can switch, and it requires two steps (deselect "any value", select "Bob (5)")

malberts:
I don't recall an e-commerce site with an "any value" option. Only a "no value" option. I remember looking at eBay with "Not Specified". Once you select any other value, the "Not Specified" option gets hidden.

My only UX concern with showing options when they might lead to redundant (`haswbfacet:P1 haswbfacet:P1=Foo`) or nonsensical (`-haswbfacet:P1 haswbfacet:P1=Foo`) queries, is that we then have a single list of (indentical looking) checkboxes where some of them behave differently. We'll end up handling these special cases in code, so it's not a technical issue, but at least on the UI there is nothing that really indicates this different behavior. Might not be a big deal though.

So if we don't hide them, then we're saying:
* these 2 special options are always visible, even when other values are ticked
* selecting one of these 2 special options triggers a "clear all"
* if one of the 2 special options were ticked, and a non-special option is selected, then it unticks the special options

PR #249, opened 2025-03-19
Show empty AND values

malberts:
For https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/180#issuecomment-2735824553

[Screencast_20250319_121944.webm](https://github.com/user-attachments/assets/9376652f-0388-49ae-a4d7-3f9ea363a9c0)

Issue #180, opened 2025-02-11
Handle OR/AND switch leading to empty results


malberts:
Multiple OR values can be used to get a non-empty search result. When switching that same filter from OR to AND, it can result in an empty search result. The same selection would not have been possible if the filter was set to AND in the first place, since the additional values would've been removed. The UI does not show the selected values, so user is now unable to easily undo the change.

https://github.com/user-attachments/assets/073bfddd-ecbb-407e-b18d-848cfaffb115

A possible solution would be to show the selected values, even when they would technically have 0 counts based on the current search. Then the user can try to deselect values until it leads to a non-empty search result.

Related: removing filters when they have no values _and no values were selected_: https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/176

JeroenDeDauw:
Is this done?

malberts:
No. https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/176 hides actually empty filters (i.e. nothing selected), and indirectly ends up hiding the non-empty filters here (i.e. something selected, but leads to empty result). We still need a way to deselect selected values where the current selection leads to an empty result.

alistair3149:
The challenge is that when the facet is in AND mode, it has no visibility of what values are in OR mode previously.
The two approaches that I can think of are either persisting the values of the last query, or run a separate query to get the values in OR mode. The second way isn't optimal and I am not sure how the first approach can be done optimally.

If there isn't a good way to handle this, I would suggest deselecting all values in the facet when clicking on the toggle. At the very least it will avoid the "impossible" query that we have with this case.

malberts:
I don't think there needs to be explicit tracking of what mode it _was_ in. 

We know what values were selected here (in `$state`):
https://github.com/ProfessionalWiki/WikibaseFacetedSearch/blob/da2a5345da91c1066b47f69f16d5510f916dd996/src/Presentation/ListFacetHtmlBuilder.php#L33-L37

My idea is if `$valueCounts` is empty AND `$state` contains selected values, then we can use a new `ZeroValueCounter` that simply returns 0 counts for a list of provided item ids.

PR #260, opened 2025-03-31
Add Any and No value filter

malberts:
For #119
For #117

[Screencast_20250331_140434.webm](https://github.com/user-attachments/assets/6556a34c-0f35-4dd0-a266-f70c1a350607)

TODO:
* [x] Add JS tests
* [x] Hide And/Or toggle when Any/No checkbox is selected


alistair3149:
I have updated the implementation to use a dropdown menu instead, and merge the AND/OR toggle into the dropdown:

https://github.com/user-attachments/assets/9661b97d-db93-46b8-a882-50193a0d86e2

JeroenDeDauw:
Does the current behavior of hiding list facets when there are no values to list always make sense?

What if I select `no value` for "field of work" and then select `no value` for "occupation", and there are no items without "occupation" that have a value for "field of work"? Would the "field of work" facet get hidden? If so, that seems a particularly bad outcome since it is still applying a filter. 

We could show the list facet always if it has a filter. If we are not already doing that.

But that might not go far enough. We are always showing the range facets, so dropping the list facets seems inconsistent. If there are no values to list, what about showing _no matching items_ instead of the list?

alistair3149:
> Does the current behavior of hiding list facets when there are no values to list always make sense?
> 
> What if I select `no value` for "field of work" and then select `no value` for "occupation", and there are no items without "occupation" that have a value for "field of work"? Would the "field of work" facet get hidden? If so, that seems a particularly bad outcome since it is still applying a filter.
> 
> We could show the list facet always if it has a filter. If we are not already doing that.
> 
> But that might not go far enough. We are always showing the range facets, so dropping the list facets seems inconsistent. If there are no values to list, what about showing _no matching items_ instead of the list?

We should always show facets that has an active filter. If we hide the facet, then there are no good way to dismiss the filter other than modifying the query string (which is not feasible).

There are multiple things that we can do to improve the experience:
- Sort the facets with active filters above the rest, and use the config as a fallback
- Add a "clear all" or "dismiss" button for each facet with an active filter
- Instead of hiding the whole facet when there are no value count, we can add an empty state stating that there are no available/matching items like you suggested

malberts:
> Does the current behavior of hiding list facets when there are no values to list always make sense?

The current idea was to hide a facet with no values when nothing was selected in that specific facet. And if values were selected in the facet, but ended up with no results due to switching from OR to AND, at least show the facet and the selected values with 0 count (TODO as per https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/180 and https://github.com/ProfessionalWiki/WikibaseFacetedSearch/pull/249).

> We are always showing the range facets

This is to some extent unintentional. Hiding this would only work if we check that all the matching results do not have that date field. Since we didn't do much with "no value" before (besides making the query work), the UI wasn't catering for it. With the list facet it was different, since we already got the allowed values from Elasticsearch, and therefore knew when nothing existed.

Additionally, with a list facet you are not supposed to select something that can lead to no results. With a range facet you can enter a range that might lead to no results. We cannot really prevent that, unless we query the min/max for that field and add validation.

In other words: range facets can also be hidden, but requires additional logic.

> what about showing no matching items instead of the list?

The current behavior is similar to e-commerce facets, and zbMath, where empty facets are not shown. But specifically: only if nothing was selected. In the example of selecting `no value` in both facets leading to no results, the TODO behavior should be to show both.

IMO adding the message would be a non-standard solution to the current problem of hiding facets inconsistently. The appeal of hiding facets is to narrow the UI and show the user exactly what can be filtered on. So if that's not entirely our intention, and we instead want to repurpose the facets to also act as "data quality" indicators (by explicitly saying when there is nothing), then I a message might be necessary.

JeroenDeDauw:
TODO tracked by https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/261

PR #280, opened 2026-01-08
Include newer PHP and MW versions in CI

alistair3149:
Also mark MW master as experimental


malberts:
master branch seems a bit useless now, because it's failing before even getting to the tests.

PR #281, opened 2026-04-14
Support external-id and edtf property types for faceted search

### LINKED ISSUE #276: Unsupported property types "external identifier" and "edtf dates"
Hello,

great plugin! We currently set up an internal Wikibase and this plugin really makes a difference in usability!

However, we've encountered a few issues and I am wondering if they are already on the agenda, as I am not sure how actively this plugin is maintained.

I think the issues are related to this issue https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/58
Namely the facetted search does not work for properties of type `external identifier` and for EDTF dates.

* For `exernal identifiers` (configured with type `list`) no search facet is shown in the UI, but if we manually include `haswbfacet:P8` or `-haswbfacet:P8` in the search query, it all of a sudden appears, but it does not work correctly.
* For EDTF dates (configured with type `range`) the search facet is shown, but does not filter correctly

We only have a few thousand test records and we checked that the index computing was finished successfully.

In our [MetaBelgica ](https://www.kbr.be/en/projects/metabelgica/)use case, we build a cultural heritage authority file, so we basically have persons with different properties for external identifiers as well as birth/death dates (EDTF) and birth/death places (items), see image below.

For us, the search ideally should support scenarios such as "Persons without identifier property X and identifier property Y" or "Persons that have a value for identifier property X and identifier property Y". Thus in a way, this issue is also related to https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/257 and it would be great if besides `list` and `range` there would be another type for properties that do not show a list (because for these identifier properties a list in the facet UI would not make much sense as these are unique values which can not be aggregated meaningfully).

We also would love to contribute directly with a pull request, but unfortunately we aren't familiar with PHP nor the internal Wikibase framework.

As a workaround for the `external identifier` issue, we duplicated each identifier property and made a regular property for it, e.g. `KBR ID` and `KBR URI`. Then we can at least search and filter with the `ID` property that is not of type external identifier.

Image

JeroenDeDauw:
Fixes https://github.com/ProfessionalWiki/WikibaseFacetedSearch/issues/276

## Summary

**Faceted search fix:** Both `external-id` and `edtf` properties use `StringValue` internally, so
the `DataValueTranslator` already handles their values. The missing piece was that these datatype IDs
were not included in the match statements of `SearchIndexFieldsBuilder` and `ListFacetQueryBuilder`,
causing their facets to be silently ignored. Range facets for EDTF are not yet supported (follow-up).

**CI fix:** Update CI for MW master compatibility — use OpenSearch 1.3 instead of deprecated
Elasticsearch 7.10, handle removal of `tests/phpunit/phpunit.php`, download `phpunit.xml.template`
excluded from tarballs, and stop using removed `SpecialPageTestBase`. MW master remains experimental
due to upstream deprecation warnings in `RevisionStore`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,34 +16,40 @@ jobs:
             php: '8.1'
             experimental: false
             coverage: false
+            search-image: 'docker-registry.wikimedia.org/repos/search-platform/cirrussearch-elasticsearch-image:v7.10.2-12'
           - mw: 'REL1_43'
             php: '8.2'
             experimental: false
             coverage: true
+            search-image: 'docker-registry.wikimedia.org/repos/search-platform/cirrussearch-elasticsearch-image:v7.10.2-12'
           - mw: 'REL1_44'
             php: '8.3'
             experimental: false
             coverage: false
+            search-image: 'docker-registry.wikimedia.org/repos/search-platform/cirrussearch-elasticsearch-image:v7.10.2-12'
           - mw: 'REL1_45'
             php: '8.4'
             experimental: false
             coverage: false
+            search-image: 'docker-registry.wikimedia.org/repos/search-platform/cirrussearch-elasticsearch-image:v7.10.2-12'
           - mw: 'master'
             php: '8.5'
-            experimental: true
+            experimental: false
             coverage: false
+            search-image: 'docker-registry.wikimedia.org/repos/search-platform/cirrussearch-opensearch-image:v1.3.20-12'
 
     runs-on: ubuntu-latest
 
     services:
       elasticsearch:
-        image: docker-registry.wikimedia.org/releng/cirrus-elasticsearch:7.10.2-s0 # Compatibility info at https://www.mediawiki.org/wiki/Extension:CirrusSearch
+        image: ${{ matrix.search-image }}
         ports:
           - 9200:9200
           - 9300:9300
         env:
           discovery.type: single-node
           ES_JAVA_OPTS: "-Xms512m -Xmx512m -Dlog4j2.formatMsgNoLookups=true"
+          OPENSEARCH_JAVA_OPTS: "-Xms512m -Xmx512m"
 
     defaults:
       run:
@@ -65,7 +71,7 @@ jobs:
             mediawiki
             !mediawiki/extensions/
             !mediawiki/vendor/
-          key: mw_${{ matrix.mw }}-php${{ matrix.php }}_v2
+          key: mw_${{ matrix.mw }}-php${{ matrix.php }}_v3
 
       - name: Cache Composer cache
         uses: actions/cache@v4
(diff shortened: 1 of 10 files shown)


JeroenDeDauw:
@SvenLieber @physikerwelt could you give this a try and see if it works well for you?

physikerwelt:
@JeroenDeDauw would you expect that we have to recreate the index to test this?
```
php maintenance/run.php CirrusSearch:UpdateSearchIndexConfig
Inferring index identifier...my_wiki_content_first
Validating number of shards...ok
Validating replica range...is 0-0 but should be 0-2...corrected
Validating shard allocation settings...done
Validating max shards per node...ok
Validating analyzers...cannot correct
<Error, collected 1 message(s) on the way, no value set>
+----------+---------------------------+--------------------------------------+
| error | rawmessage | This script encountered an index dif |
| | | ference that requires that the index |
| | | be
copied, indexed to, and... |
+----------+---------------------------+--------------------------------------+
```

physikerwelt:
Screenshot 2026-04-17 at 05 23 25
I didn't want to blow away the search index in production, but in our staging server, the patch worked (after recreating the search index).

JeroenDeDauw:
@physikerwelt yes, you'll have to rebuild, or wait for values to slowly appear as their indexing is triggered by edits to items. 

Thanks for testing. Looks like you verified external-id works. What about EDTF?

physikerwelt:
Sorry, I did not test that.

physikerwelt:
The recreation of the index makes testing (in our environment) a bit painful. 
Do you think we could relax
```
php maintenance/run.php CirrusSearch:UpdateSearchIndexConfig 
```
to
```
php maintenance/run.php CirrusSearch:UpdateSearchIndexConfig --reindexAndRemoveOk --indexIdentifier now
```
?

JeroenDeDauw:
No idea

gcgbarbosa:
@JeroenDeDauw, 

Thanks a lot for adding this functionality.
I've tested the PR.

Both `EDTF` and `External identifier` work.

The only issue is, for EDTF, range search does not work.

I assumed this was expected, since you've mentioned in the commit that:

```text
Both external-id and edtf properties use StringValue internally
```

JeroenDeDauw:
That's right, the EDTF string are stored. Would it be better to store a numeric timestamp so you can do range searches?

We do that in Wikibase RDF for EDTF values: we turn the EDTF into a standard timestamp, but the downside there is that information and precision are lost. It's especially awkward for ranges. For the RDF can we solved that by exporting both the lower and upper bound of a range.

physikerwelt:
Maybe the improvement of EDTF can be a seperate issue? When working with BibTeX data, you become used to queries like `publicationYear=2000 or publicationYear=2001 or ...`. You think I should really improve the datamodel, but it works for the time being.

SvenLieber:
@JeroenDeDauw Thanks for addressing the issue, @gcgbarbosa verified that it works for EDTF and external identifiers. I agree with @physikerwelt , the range query can be another issue if it is not a priority for them. This can then be addressed in the future if it becomes a priority for one of the customers.

JeroenDeDauw:
A follow-up issue for an EDTF-specific UI that allows range queries seems sensible, yes. Feel free to file it.

